PR #1, opened 2015-06-19
Remove some `elses`

CentaurWarchief:
I'm just applying some kind of [object calisthenics](http://www.bennadel.com/blog/2375-object-calisthenics-in-javascript-my-first-attempt.htm) here.

### DIFF
--- a/src/core/CSSCore.js
+++ b/src/core/CSSCore.js
@@ -36,14 +36,17 @@ var CSSCore = {
       'multiple classes.', className
     );
 
-    if (className) {
-      if (element.classList) {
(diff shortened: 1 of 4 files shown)


zpao:
Some of this is fine but as of right now we're not ready to take PRs here. We'll get our tooling into better shape over the next month or so but until then, if it's not a bug fix, we're not to merge any PRs.

zpao:
Thanks btw, I appreciate the initiative. Come back in a little while when we're all settled in :)

CentaurWarchief:
@zpao Aw, gotcha!

:+1:

PR #3, opened 2015-07-22
Add "repository" to package.json

yungsters:
Silences this warning when running `npm install`:

```
npm WARN package.json fbjs@0.1.0-alpha.1 No repository field.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "0.1.0-alpha.1",
   "description": "",
   "main": "index.js",
+  "repository": "facebook/fbjs",
   "scripts": {
     "test": "NODE_ENV=test jest",
     "build": "gulp build",


package.json
```diff
@@ -3,6 +3,10 @@
   "version": "0.1.0-alpha.1",
   "description": "",
   "main": "index.js",
+  "repository": {
```
zpao: Let's be fancy and just use `"repository": "facebook/fbjs"` since it works :). As you have (and we have in React too I think), it might not be quite right - https://docs.npmjs.com/files/package.json#repository

yungsters: Done.

zpao: Oh no, I meant you can get rid of the object entirely and the value is just a string :) "facebook/fbjs" is understood to be git by default.

zpao:
👍🏻

PR #4, opened 2015-07-22
Sync some more dependencies out

yungsters:
Adds a bunch of dependencies (and moves `mapObject.js`) that will be needed by future projects.

**Test Plan:**

 - `npm install`
 - `npm test`


zpao:
👍🏻

PR #5, opened 2015-07-22
Add Deferred and PromiseMap dependencies

yungsters:
Pulls in `Deferred` and `PromiseMap`. This also required adding `promise` as a package dependency and adding it to the `moduleMap` config.

**Test Plan:**

```
npm test
gulp
```

### DIFF
--- a/src/core/Deferred.js
+++ b/src/core/Deferred.js
@@ -1,0 +1,66 @@
+/**
+ * Copyright 2013-2015, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @providesModule Deferred
+ * @typechecks
+ * @flow
+ */
+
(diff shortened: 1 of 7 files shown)


yungsters:
We should try to consolidate the Babel configuration that currently exists in `preprocessor.js` and `gulpfile.js`.

yungsters:
Actually, Flow won't be happy if we don't have a real `Promise` module, so let me fix this.

src/core/Deferred.js
```diff
@@ -0,0 +1,61 @@
+/**
+ * Copyright 2004-present Facebook. All Rights Reserved.
```
zpao: We should make sure this header is updated

zpao:
> We should try to consolidate the Babel configuration that currently exists in preprocessor.js and gulpfile.js.

[My comment](https://github.com/facebook/fbjs/blob/master/scripts/jest/preprocessor.js#L1-L2) agrees with you :)

zpao:
accepting because I know you'll fix the license headers

yungsters:
> accepting because I know you'll fix the license headers

So much faith. ;]

PR #7, opened 2015-07-22
Replace Promise => promise module mapping with Promise module

yungsters:
Instead of using `_moduleMap`, this adds a `Promise` module that simply exports `promise`.

Ideally, we would just use `_moduleMap` and avoid the (tiny) overhead of having a `Promise` module. However, this is insufficient for Flow which processes the files pre-transpilation. Without this, `flow check` would complain that it cannot find a module named `Promise`.


zpao:
👍🏻

PR #12, opened 2015-07-23
Match \r or \r\n linebreaks when scanning for @providesModule

steveluscher:
My module-map.json contained only an empty object, post-build, on my machine. Making the `@providesModule` regex less restrictive with respect to line breaks fixed it.


zpao:
:thumbsup: I even fixed the same thing a long time ago in haste docblock parsing… Thanks!

PR #15, opened 2015-07-23
Sync all the things!

yungsters:
This is using a new script to programmatically (and uniformly and systematically) sync all the modules in fbjs with the internal versions.


src/core/getMarkupWrap.js
```diff
@@ -65,19 +65,15 @@ var svgElements = [
...
-  'mask',
   'path',
-  'pattern',
   'polygon',
   'polyline',
   'radialGradient',
   'rect',
   'stop',
   'text',
-  'tspan',
```
zpao: We actually want these.

zpao:
:thumbsup:

PR #17, opened 2015-07-23
Include flow comments in compiled output

zpao:
This should make it such that we can properly run flow in projects consuming fbjs.

(Sorry for the package.json noise, npm3 sorts on writing)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -6,6 +6,7 @@ var runSequence = require('run-sequence');
 
 var babelPluginDEV = require('./scripts/babel/dev-expression');
 var babelPluginRequires = require('./scripts/babel/rewrite-requires');
+var babelPluginFlowComments = require('babel-plugin-flow-comments');
 var gulpModuleMap = require('./scripts/gulp/module-map.js');
 
 var paths = {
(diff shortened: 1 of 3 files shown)


zpao:
Well this can't be right… that diff definitely has some syntax errors. https://gist.github.com/zpao/29cc0a4bd03791d08cd1

I'll make this so that we run tests with this transform and catch things like that.

zpao:
And now it's works because @sebmck is awesome.

yungsters:
This change causes dependent projects to suddenly have many false flow errors. A cursory investigation shows that 1) flow does not like the code generated by babel (e.g. `_createClass` assigns properties to an object without a defined shape), and 2) generic types (e.g. `Foo<T>`) are not preserved and become invalid identifiers.

zpao:
What's the best move? Revert?

sebmck:
Setting `{ "loose": "es6.classes" }` should be enough to make Flow understand the class structure. Not entirely clear on `Foo<T>` but if it's a Babel issue let me know and I'll fix it.

yungsters:
Cool, I'll try `{ "loose": "es6.classes" }`. Thanks!

I might have found another issue:

```
function everyObject(
  object: ?Object,
  callback: (value: any, name: string, object: Object) => any,
  context?: any,
): boolean {
```

...is being transpiled into...

```
function everyObject(object /*: ?Object*/, callback /*: (value: any, name: string, object: Object) => any*/, context? /*:: ?: any*/) /*: boolean*/ {
```

The problem is `context?`, which is invalid syntax.

sebmck:
That'd be because the `flow` transformer has been blacklisted which is responsible for stripping those. Fixed via babel-plugins/babel-plugin-flow-comments@df836e49e846fbec375d7f4de5f8c0a182333487 and released as of `babel-plugin-flow-comments@1.0.9`

yungsters:
Great. I'll give `babel-plugin-flow-comments@1.0.9` a spin.

I'll open an issue for the generic types on https://github.com/babel-plugins/babel-plugin-flow-comments.

yungsters:
Opened babel-plugins/babel-plugin-flow-comments#15.

PR #18, opened 2015-07-24
Unmock __forks__/fetch.js to fix tests

yungsters:
When I moved `fetch.js` from `src/fetch` into `src/__forks__`, it became unmocked. This reverses that (although the mechanism for doing so really sucks).


zpao:
:thumbsup: (I'll work on getting TravisCI set up here so we know as soon as we break tests, I noticed last night, was going to ask today)

PR #19, opened 2015-07-24
Sync out `shallowEqual` (from React)

yungsters:
Syncs `shallowEqual` into this repository. It is needed by future projects (not only `react`, which is where it currently resides).


zpao:
Cool. I don't think we ever actually shipped this in a real React release so we may need to update the react-codemod @cpojer has.

cpojer:
The codemod uses `shallowCompare`, not `shallowEqual`. But I'm still waiting for us to invent something better :)

PR #21, opened 2015-07-25
Fix Jest via Package Composition

yungsters:
When a package that composes `fbjs` runs Jest, some things do not currently work. In particular:

 - Module names are not rewritten for `require.requireActual`, `jest.dontMock`, `jest.mock`, and `jest.genMockFromModule`.
 - Modules using `Object.assign` or the spread operator fatal because it has not been polyfilled.
 - The `es6-map` polyfill pulls in `EventEmitter` which throws in the Jest environment.

Here, I address these issues by:

 - Replacing `es6-map` with `core-js`.
 - Polyfill `Object.assign` in `environment.js` via `core-js`.
 - Add rewrite transformations for the remaining module literal strings.


scripts/babel/rewrite-modules.js
```diff
@@ -0,0 +1,123 @@
...
+          [t.literal(module)]
+        );
+      }
+    }
+  }
+
+  /**
+   * Transforms `require.requireActual('Foo')`.
+   */
+  function transformRequireActualCall(context, call) {
```
clentfort: This function seems unused.

scripts/babel/rewrite-modules.js
```diff
@@ -0,0 +1,123 @@
...
+  if (process.env.NODE_ENV !== 'test') {
+    return './' + module;
+  }
+}
+
+module.exports = function(babel) {
+  var t = babel.types;
+
+  /**
+   * Transforms `require('Foo')` and `require.requireActual('Foo')`.
```
clentfort: Could you explain to what they are transformed?

scripts/babel/rewrite-modules.js
```diff
@@ -0,0 +1,123 @@
...
+function mapModule(context, module) {
+  // Look in options passed to babel for `_moduleMap` which contains a map of
+  // replacements. This allows other packages to be published to npm and used
+  // directly without being a part of the same build.
+  var moduleMap = context.state.opts._moduleMap || {};
+  if (moduleMap.hasOwnProperty(module)) {
+    return moduleMap[module];
+  }
+  // Jest understands the haste module system, so leave modules intact.
+  if (process.env.NODE_ENV !== 'test') {
```
clentfort: Is there by any chance a better way to detect a jest environment? This looks like it might lock project that uses fbjs into jest for testing (since `require` calls wont be transformed in any test environment).

zpao: We can set whatever environment var we want when running jest - we could set NODE_ENV=jest.

FWIW, this whole setup is very much tuned for FB and our needs. Some random project using fbjs doesn't need to use our babel transforms. The goal is to make it easier for other FB projects with the same needs to use the same code.

scripts/babel/rewrite-modules.js
```diff
@@ -0,0 +1,123 @@
...
+        t.isIdentifier(call.callee.property, {name: 'requireActual'})
+      )
+    ) {
+      return;
+    }
+    var moduleArg = call.arguments[0];
+    if (moduleArg && moduleArg.type === 'Literal') {
+      var module = mapModule(context, moduleArg.value);
+      if (module) {
+        return t.callExpression(
```
zpao: Does this work for requireActual? Or does this fall into the same situation where we need to create a MemberExpression?

scripts/babel/rewrite-modules.js
yungsters: Oops, good catch. I originally started doing this separate from `require('Foo')`, but it turned out to be almost the same transform.

scripts/babel/rewrite-modules.js
yungsters: :+1:

scripts/babel/rewrite-modules.js
yungsters: We can re-use the MemberExpression for `require.requireActual`.

This is different from Jest because the `jest` methods can be chained. If chained, the `member.object` has to be point at a new CallExpression. (However, like here, we're still able to re-use the `jest` identifier.)

yungsters:
Is this good to go?

scripts/jest/environment.js
```diff
@@ -1 +1,3 @@
+require('core-js/modules/es6.object.assign');
```
zpao: Just to be clear, this one is writing to the global `Object` but the `Map` one isn't?

yungsters: @zpao The `core-js/library` modules avoid global namespace pollution. I found `core-js/modules` from snooping around, but now I think I might be wrong. Let me rtfm and get back to you.

PR #24, opened 2015-07-30
Flow: build includes instead of comment annotations

### LINKED ISSUE #23: npm'd version fails flow
The transpiled version has stuff like:

```js
function Deferred /*:: <Tvalue, Treason>*/() {
  // stuff ...
}

Deferred.prototype.resolve = function resolve(value /*: Tvalue*/) /*: void*/ {
  this._settled = true;
  this._resolve(value);
};
```

Which makes `Tvalue` undefined, b/c it's not in the same scope.
Not sure if this is a flow bug or not, but it makes it so I can't use this lib :(

zpao:
Fixes #23

This should let us do the following in .flowconfigs:

```ini
[include]
node_modules/fbjs/flow/include

[libs]
node_modules/fbjs/flow/lib
```


yungsters:
Niiice, thanks!

PR #27, opened 2015-08-06
Build manual mocks into lib/__mocks__

yungsters:
Exports manual mocks so that composite projects can run tests using them.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -43,7 +43,7 @@ var moduleMapOpts = {
 };
 
 gulp.task('clean', function(cb) {
-  del([paths.lib.dest, paths.flowInclude], cb);
+  del([paths.lib.dest, paths.mocks.dest, paths.flowInclude], cb);
 });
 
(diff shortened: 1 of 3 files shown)


yungsters:
There is one more outstanding issue. Mocks that use `jest.genMockFromModule('X')` will be transformed to be `jest.genMockFromModule('./X')`. This actually needs to be `jest.genMockFromModule('../X')` since `__mocks__` is an extra level deep.

zpao:
:thumbsup: I'll see if I can make that work

gulpfile.js
```diff
@@ -31,24 +43,32 @@ var moduleMapOpts = {
 };
 
 gulp.task('clean', function(cb) {
-  del([paths.lib, paths.flowInclude], cb);
+  del([paths.lib.dest, paths.flowInclude], cb);
```
josephsavona: include `paths.mocks.dest` for completeness, in case it diverges?

PR #28, opened 2015-08-06
Add manual mock for ErrorUtils

yungsters:


src/stubs/__mocks__/ErrorUtils.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * Copyright 2004-present Facebook. All Rights Reserved.
```
zpao: Let's fix this

PR #32, opened 2015-08-13
Respect the empty string as a module rewrite prefix

### LINKED ISSUE #31: Rewrite module and _modulePrefix
I know this is a library initially for Facebook internal projects but instead of copying the script, I was trying to reuse what you did here and noticed something in the script `rewrite-modules`.

The default value for `_modulePrefix` is `./`.
```
var modulePrefix = context.state.opts._modulePrefix || './';
```
I think the default value should be an empty string, because if you require an external module from your node_modules folder, you don't want to prefix it with `./`.

For my needs, and because my project uses external modules, I would need the default to be an empty string, but because of the line above, you'll understand it's not possible.

If for your needs the default should stay `./`, would at least be possible to check if the option is set?

zpao:
Fixes #31

This still allows other crappy prefixes like `false`. We could probably do a `typeof === 'string'` check instead…

cc @yungsters

### DIFF
--- a/scripts/babel/rewrite-modules.js
+++ b/scripts/babel/rewrite-modules.js
@@ -21,7 +21,10 @@ function mapModule(context, module) {
   }
   // Jest understands the haste module system, so leave modules intact.
   if (process.env.NODE_ENV !== 'test') {
-    var modulePrefix = context.state.opts._modulePrefix || './';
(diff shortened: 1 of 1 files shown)


yungsters:
:+1:

PR #33, opened 2015-08-14
Make `fetchWithRetries` configs optional

yungsters:
Changes the Flow annotations for `fetchWithRetries` to make it easier to compose.


zpao:
Want to merge #30 and drop your first commit ;)

yungsters:
@zpao Oh, I didn't see that. Good catch!

zpao:
👍

Issue #36, opened 2015-08-16
Passing build & travis setup


glenjamin:
Checking this out locally, installing deps & trying to run the tests I get the following failure (node 0.10.40):

```
 FAIL  src/fetch/__tests__/fetchWithRetries-test.js (0.639s)
● fetchWithRetries › it defaults fetch timeout to 15s
  - Expected Function to be called.
        at /Users/glen/Development/Github/fbjs/src/fetch/__tests__/fetchWithRetries-test.js:114:26
        at mockConstructor (/Users/glen/Development/Github/fbjs/node_modules/jest-cli/src/lib/moduleMocker.js:144:27)
        at undefined (eval at makeComponent (/Users/glen/Development/Github/fbjs/node_modules/jest-cli/src/lib/moduleMocker.js:157:15), <anonymous>:2:52)
        at _timers.(anonymous function).callback (/Users/glen/Development/Github/fbjs/node_modules/jest-cli/src/lib/FakeTimers.js:413:23)
        at FakeTimers._runTimerHandle (/Users/glen/Development/Github/fbjs/node_modules/jest-cli/src/lib/FakeTimers.js:450:7)
        at FakeTimers.runAllTimers (/Users/glen/Development/Github/fbjs/node_modules/jest-cli/src/lib/FakeTimers.js:170:10)
        at Loader.jestRuntime.exports.runAllTimers (/Users/glen/Development/Github/fbjs/node_modules/jest-cli/src/HasteModuleLoader/HasteModuleLoader.js:1039:42)
        at Spec.<anonymous> (/Users/glen/Development/Github/fbjs/src/fetch/__tests__/fetchWithRetries-test.js:116:10)
        at Timer.listOnTimeout [as ontimeout] (timers.js:121:15)
● fetchWithRetries › it preserves fetch timeout of 0s
  - Expected Function to be called.
        at /Users/glen/Development/Github/fbjs/src/fetch/__tests__/fetchWithRetries-test.js:129:26
        at mockConstructor (/Users/glen/Development/Github/fbjs/node_modules/jest-cli/src/lib/moduleMocker.js:144:27)
        at undefined (eval at makeComponent (/Users/glen/Development/Github/fbjs/node_modules/jest-cli/src/lib/moduleMocker.js:157:15), <anonymous>:2:52)
        at _timers.(anonymous function).callback (/Users/glen/Development/Github/fbjs/node_modules/jest-cli/src/lib/FakeTimers.js:413:23)
[...]

PR #37, opened 2015-08-18
Disable broken tests in fetchWithRetries-test.js

yungsters:
Ugh... not sure what the deal is with Jest and `promise`, but going to just disable these tests for now.

Originally reported by @glenjamin via #36.

cc @cpojer

### DIFF
--- a/scripts/jest/preprocessor.js
+++ b/scripts/jest/preprocessor.js
@@ -2,7 +2,9 @@ var assign = require('object-assign');
 var babel = require('babel');
 var babelDefaultOptions = require('../babel/default-options');
 
-var babelOpts = babelDefaultOptions;
+var babelOpts = assign({}, babelDefaultOptions, {
+  retainLines: true,
+});
 
 module.exports = {
   process: function(src, path) {
--- a/src/fetch/__tests__/fetchWithRetries-test.js
+++ b/src/fetch/__tests__/fetchWithRetries-test.js
@@ -114,7 +114,8 @@ describe('fetchWithRetries', () => {
     ));
   });
 
-  it('defaults fetch timeout to 15s', () => {
+  // Test fails when used with npm `promise` due to Jest timing issues.
+  xit('defaults fetch timeout to 15s', () => {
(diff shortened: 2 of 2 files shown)


scripts/jest/preprocessor.js
```diff
@@ -2,7 +2,9 @@ var assign = require('object-assign');
 var babel = require('babel');
 var babelDefaultOptions = require('../babel/default-options');
 
-var babelOpts = babelDefaultOptions;
+var babelOpts = assign({}, babelDefaultOptions, {
+  retainLines: true,
```
yungsters: Also added this for accurate stack trace line numbers. We do this in **facebook/relay**.

cpojer:
this is weird, they were definitely working when I upgraded to jest 0.5. The bug was reported with node 0.10 which is not supported any more when attempting to run the tests for Relay. If this works with io + jest 0.5 I wouldn't merge this PR.

glenjamin:
Is it possible to keep the build running on 0.10, 0.12 & iojs?

I would expect 0.10 and 0.12 to still make up the majority of users

yungsters:
The bug still exists for jest-cli@0.5.0 on iojs@3.0.0. I also tried using the native `Promise` instead of the npm `promise` with no luck. Going to proceed with this for now.

@glenjamin Changing this to use jest-cli@0.5.0 only means that iojs is needed as a dev dependency. It shouldn't change anything else about this package or any dependent packages.

cpojer:
mhh, ok. Timers are weird.

glenjamin:
@yungsters sorry, I was referring to the hypothetical travis CI build - AIUI if jest-cli only runs on iojs, then the tests cannot run on 0.10 or 0.12 and the test suite provides no protection from regressions on these platforms?

zpao:
@glenjamin yea… it's not the best situation. I don't know the best way to actually make that work with jest. We could potentially do something to explicitly install jest @0.4 in a node 0.10 env (0.12 pretty much just won't work).

PR #38, opened 2015-08-18
Upgrade to jest-cli@5.0.0

yungsters:


package.json
```diff
@@ -18,12 +18,15 @@
...
     "gulp-util": "^3.0.4",
-    "jest-cli": "^0.4.5",
+    "jest-cli": "^0.5.0",
     "merge-stream": "^0.1.8",
     "object-assign": "^3.0.0",
     "run-sequence": "^1.1.0",
     "through2": "^2.0.0"
   },
+  "engines": {
+    "node": ">=2"
```
zpao: We really shouldn't be doing - this flag is only consumed in downstream installs. Saying that fbjs is only compatible with node > 2 is a lie. It might be all we test with but that doesn't make it true.

yungsters: Should we remove it from **facebook/react-relay**, too?

zpao: Probably. It would be good to have some prevention mechanism to make sure the right versions of tools are being used for development (Eg, check node and npm versions)

zpao: Actually 1 place that might use this info is TravisCI. Just confirm in relay what version gets run both before and with a PR to remove this line.

cpojer: Travis doesn't look at this. See https://github.com/facebook/jest/blob/master/.travis.yml

fbjs doesn't use travis, care to set that up? :P

zpao: I will eventually :P We're not taking PRs here right now so it's not a huge deal yet. I meant for Relay and React where we care.

PR #39, opened 2015-08-18
Remove `engines` from package.json

yungsters:
cc @zpao @cpojer


zpao:
:thumbsup: you so fast

PR #40, opened 2015-08-19
Remove `../flow` from .flowconfig

yungsters:
The `flow/include` directory only contains output for consumers of **facebook/fbjs**. These double up on all the definitions in `src/`.

Fixing this fixes all the issues when running `flow check src`.

### DIFF
--- a/src/.flowconfig
+++ b/src/.flowconfig
@@ -5,7 +5,7 @@
 ../node_modules/promise
 
 [libs]
-../flow
+../flow/lib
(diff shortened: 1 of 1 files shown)


zpao:
There's also `lib/dev.js`, do we not need that here?

yungsters:
That's what I get for hiding all `lib` folders in Sublime Text.

zpao:
:) Thanks!

PR #43, opened 2015-08-19
Split scripts into standalone package

### LINKED ISSUE #34: Split out scripts into seprate package
I want to split out the scripts and ship them as their own package. We can keep the `fbjs` package focused the code (and flow integration). The scripts package would contain the babel transforms, gulp scripts, future lint stuff. We're already not including the deps right for fbjs (eg, missing deps for gulp scripts) and I don't think `gulp-utils` should end up getting installed when you use React or Relay.

zpao:
Not quite done but gets the plan across. The problem I'm trying to solve here is make it possible to ship more and more support scripts without bloating the production dependencies (eg, lint is coming). Everything I moved from `devDependencies` into fbjs-scripts actually should have been a regular dependency of fbjs (but wasn't so using some scripts was actually broken). The flow stuff sits somewhere in between so I just left it as something that gets shipped with fbjs.

This uses `file:scripts` to pull in `fbjs-scripts` as a package without having to hit npm. This will also let us make changes to fbjs tooling without having to ship it to npm to actually use it, then we can ship it when it's ready. Changes will need 'npm install ./scripts' to be run (since it's copied, but I think this might be more sustainable than linking).

Any thoughts on this @yungsters, @steveluscher?

Fixes #34

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -1,0 +1,17 @@
+# fbjs-scripts
+
+This is a collection of tools and scripts intended to be used in conjunction with `fbjs`. Previously these were shipped as a part of `fbjs`.
+
+```js
+// before (fbjs@0.1.0)
+var invariant = require('fbjs/lib/invariant');
+var devExpression = require('fbjs/scripts/babel/dev-expression');
+
+// after (fbjs, fbjs-scripts@0.2.0)
+var invariant = require('fbjs/lib/invariant');
+var devExpression = require('fbjs-scripts/babel/dev-expression');
+```
+
+## Why?
+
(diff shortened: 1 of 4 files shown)


steveluscher:
That sounds like a good line to draw! Do it.

scripts/README.md
```diff
@@ -0,0 +1,17 @@
+# fbjs-scripts
+
+This is a collection of tools and scripts intended to be used in conjunction with `fbjs`. Previously these were shipped as a part of `fbjs`.
+
+```js
+// before (fbjs@0.1.0)
```
yungsters: When were we at 0.1.0? Haha.

yungsters:
Sounds good to me.

PR #45, opened 2015-08-28
Enable inline-require transform for tests.

cpojer:
Relay diff incoming too.

### DIFF
--- a/scripts/babel/inline-requires.js
+++ b/scripts/babel/inline-requires.js
@@ -1,0 +1,132 @@
+/**
+ * Copyright 2013-2015, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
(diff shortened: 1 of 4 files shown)


cpojer:
See https://github.com/facebook/relay/pull/187

scripts/babel/inline-requires.js
```diff
@@ -0,0 +1,128 @@
+/**
+ * Copyright 2004-present Facebook. All Rights Reserved.
```
zpao: fixit

scripts/babel/inline-requires.js
```diff
@@ -0,0 +1,128 @@
...
+ */
+
+'use strict';
+/*jslint node: true */
```
zpao: rm

zpao:
:thumbsup:

PR #50, opened 2015-09-08
Add functional/filterObject

yungsters:
Also, syncing a little change from internal.


yungsters:
cc @josephsavona @cpojer

josephsavona:
:+1: @yungsters - let's bump the package version & publish so we can update the dependency in `react-relay`?

PR #51, opened 2015-09-17
Ignore `config` option when creating a cache key.

cpojer:
We don't need it because it can't influence the output of a test file. It is also not being stringified consistently and results in more cache misses than necessary.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "whatwg-fetch": "^0.9.0"
   },
   "devEngines": {
-    "node": "3.x",
+    "node": ">=3",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -61,7 +61,7 @@
     "whatwg-fetch": "^0.9.0"
   },
   "devEngines": {
-    "node": "3.x",
+    "node": ">=3.x",
```
zpao: `>=3` should be fine (mixing versioning with `.x` and `>=` is gross)

cpojer: done!

PR #52, opened 2015-09-19
[fetch] Correctly reject final try on error

clentfort:
Correctly reject the promise if the final `fetch`-attempt returned with an HTTP
status not between 200 and 299. The error will have an extra property `response`
which stores the response `fetch` returned.

Fixes https://github.com/facebook/relay/issues/347

### DIFF
--- a/src/fetch/fetchWithRetries.js
+++ b/src/fetch/fetchWithRetries.js
@@ -86,7 +86,12 @@ function fetchWithRetries(
             retryRequest();
           } else {
             // Request was not successful, giving up.
-            reject(response);
+            var error: any = new Error(sprintf(
+                'fetchWithRetries(): Got an error as a final response from ' +
+                'server, tried %s times.',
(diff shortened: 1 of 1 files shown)


src/fetch/fetchWithRetries.js
```diff
@@ -86,7 +86,12 @@ function fetchWithRetries(
             retryRequest();
           } else {
             // Request was not successful, giving up.
-            reject(response);
+            var error: any = new Error(sprintf(
```
clentfort: Cast to `any` so I can assign the `response` property.

zpao:
cc @yungsters who can probably take a look tomorrow. I'll let him decide which PR is the preferred one.

yungsters:
I prefer #53 without the JSON parsing, mainly because it has unit tests.

clentfort:
The other PR is to prefer.

PR #53, opened 2015-09-20
Ensure fetchWithRetries is rejected with an Error

emmenko:
After the last retry fails, the Promise should be rejected with an `Error` instead of the fetch `Response` object.
This will parse the body and append it to the returned `Error` object.

Fixes https://github.com/facebook/relay/issues/347

### DIFF
--- a/src/fetch/__tests__/fetchWithRetries-test.js
+++ b/src/fetch/__tests__/fetchWithRetries-test.js
@@ -89,6 +89,34 @@ describe('fetchWithRetries', () => {
     expect(handleNext).toBeCalledWith(successfulResponse);
   });
 
+  it('gives up if response failed after retries', () => {
+    var init = {retryDelays: [600]};
+    var failedResponse = mockResponse(500);
+    var handleCatch = jest.genMockFunction();
+    fetchWithRetries('https://localhost', init)
+      .then(handleNext).catch(handleCatch);
+    expect(fetch.mock.calls.length).toBe(1);
+    fetch.mock.deferreds[0].resolve(failedResponse);
+    for (var ii = 0; ii < 100; ii++) {
+      if (fetch.mock.calls.length < 2) {
+        jest.runOnlyPendingTimers();
+      } else {
+        break;
+      }
+    }
+    // Resolved with `failedResponse`, next run is scheduled
+    expect(fetch.mock.calls.length).toBe(2);
+    fetch.mock.deferreds[1].resolve(failedResponse);
+    // No more re-tries, it should reject with an `Error`
+    expect(handleNext).not.toBeCalled();
+    jest.runAllTimers();
+    var errorArg = handleCatch.mock.calls[0][0];
+    expect(errorArg instanceof Error).toBe(true);
+    expect(errorArg.message).toEqual('fetchWithRetries(): Still no ' +
+      'successful response after 2 retries, giving up.');
+    expect(errorArg.response).toEqual(failedResponse);
+  });
+
   it('retries the request if the previous attempt timed-out', () => {
     var retries;
     var retryDelays = [1000, 3000];
--- a/src/fetch/fetchWithRetries.js
+++ b/src/fetch/fetchWithRetries.js
@@ -86,7 +86,13 @@ function fetchWithRetries(
             retryRequest();
           } else {
             // Request was not successful, giving up.
-            reject(response);
+            var error: any = new Error(sprintf(
+              'fetchWithRetries(): Still no successful response after ' +
+              '%s retries, giving up.',
+              requestsAttempted
+            ));
+            error.response = response;
+            reject(error);
           }
         }
       }).catch(error => {


emmenko:
I see there is another PR for this issue https://github.com/facebook/fbjs/pull/52

Not sure what you guys prefer :)

clentfort:
IHMO its not necessary to try to parse the body. In fact it would be better to just include the entire response in the error. My PR has no unit-tests so this one is to preferred over mine.

emmenko:
@clentfort maybe it's not necessary but I'm wondering about this use case: when using the `renderFailure` callback I want to inspect / render the JSON response body. So if the `error` doesn't have the body parsed I'm not sure how I can render the component after resolving the promise. 
See https://github.com/facebook/relay/issues/347

```js
renderFailure={function(error, retry) {
  return (
    <div>
      <p>{error.body}</p>
      <p><button onClick={retry}>Retry?</button></p>
    </div>
  );
}}
```

clentfort:
Do the parsing of the body in the `renderFailure`? I think it makes more sense to keep it generic so it can easily be reused in libraries to come. Option value

emmenko:
But when I tried that I got an error (the `error` in this case is the `Response` object). Maybe I'm doing something wrong...any idea?

```js
<Relay.RootContainer
  Component={Home}
  route={homeQueries}
  renderFailure={(error, retry) => {
    error.json().then(result => {
      return (
        <div>
          <p>{result.errors}</p>
          <p><button onClick={retry}>Retry?</button></p>
        </div>
      )
    })
  }}
/>
```

screen shot 2015-09-19 at 00 48 49

yungsters:
> the `error` in this case is the `Response` object

@emmenko Shouldn't `error` always be an `Error` object?

I agree with @clentfort's suggestion to continue with your fix (to `reject` with an `Error` object instead of the response) but not to try JSON parsing the response before assigning it to `error.body`. If for no other reason than ambiguity:

```
renderFailure={error => {
  var {body} = error;
  if (typeof body === 'string') {
    // Is `body` a string because the response was a JSON-serialized string?
    // Or did `body` fail to JSON parse?
  }
}}
```

emmenko:
> Shouldn't error always be an `Error` object?

Yes of course, it was just the example sorry. Image it is `error.body.json()` or `error.response.json()`.

> but not to try JSON parsing the response before assigning it to error.body

Again, the only problem I see is that if I return an `Error` without parsing the `Response` I cannot do it anymore in the `renderFailure` function. Afaics, `renderFailure` expects to return a react element synchronously.
https://github.com/facebook/relay/blob/70c12c5187a1cfb323b3f09649aa1be9f7f3c32e/src/container/RelayRootContainer.js#L43
https://github.com/relay-tools/react-router-relay/blob/master/src/Container.js#L45

So there is no way to parse the response body in the `renderFailure` function.

Maybe there are some reasons for that that I'm not aware of, or maybe there is another way to do it. 
Regardless, I'd like to clarify this first, thanks :)

emmenko:
PS: I'm not saying my solution is the correct one, I just want to understand and find the proper solution for parsing the response :)

yungsters:
Ahh, I see what you mean now. That makes a lot of sense when used with Relay, but the concern probably does not belong in `fetchWithRetries`.

If we wanted to make the assumption that responses for server failures (that trigger retries) are JSON strings, we should do so in `RelayDefaultNetworkLayer`. Notice that this would mirror how `RelayDefaultNetworkLayer` is responsible for calling `response.json()` before resolving the `RelayQueryRequest` promise.

By keeping `fetchWithRetries` generic, we also make it more useful for other custom network layers that may not make the same assumptions.

Also, if we don't try to `text()`/`json()`-ify the response, we should probably rename the key to `response`.

emmenko:
> we should do so in `RelayDefaultNetworkLayer`

I see, that makes a lot of sense either :)

Yes, I agree we should keep this generic then :+1:  
I'll provide the changes and squash the commits. 

PS: proper error handling then should be done where? [Here](https://github.com/facebook/relay/blob/3e7738a8ec2b6a02a208f19a2708102a421ade6a/src/network-layer/default/RelayDefaultNetworkLayer.js#L91) or in some other places?

yungsters:
Yeah, [that](https://github.com/facebook/relay/blob/3e7738a8ec2b6a02a208f19a2708102a421ade6a/src/network-layer/default/RelayDefaultNetworkLayer.js#L91) seems like the right place.

emmenko:
:+1: 

So should I open an issue or do you guys take care of that?

emmenko:
@yungsters here we go, I've squashed the commits. Please have another look.

src/fetch/fetchWithRetries.js
```diff
@@ -86,7 +86,12 @@ function fetchWithRetries(
             retryRequest();
           } else {
             // Request was not successful, giving up.
-            reject(response);
+            var error: any = new Error(sprintf(
+              'fetchWithRetries(): Gave up after %s re-tries.',
```
clentfort: Maybe we can tune the error message a bit. This is the exact same error message one gets after the last attempted timed-out. In this case we tried X times and the final attempted returned an error status code.

emmenko: You mean something like `fetchWithRetries(): Still no successful response after %s re-tries, giving up.` ?

clentfort: Yup! :) A unique error message makes it easier to debug user reported errors.

emmenko: Done, thanks! :)

src/fetch/fetchWithRetries.js
```diff
@@ -86,7 +86,12 @@ function fetchWithRetries(
             retryRequest();
           } else {
             // Request was not successful, giving up.
-            reject(response);
+            var error: any = new Error(sprintf(
+              'fetchWithRetries(): Still no successful response after ' +
+              '%s re-tries, giving up', requestsAttempted
```
zpao: nit: "retries"

src/fetch/__tests__/fetchWithRetries-test.js
```diff
@@ -89,6 +89,34 @@ describe('fetchWithRetries', () => {
     expect(handleNext).toBeCalledWith(successfulResponse);
   });
 
+  it('gives up if response failed after retries', () => {
+    var init = {retryDelays: [600]};
+    var failedResponse = mockResponse(500);
+    var handleCatch = jest.genMockFunction();
+    fetchWithRetries('https://localhost', init)
+    .then(handleNext).catch(handleCatch);
```
zpao: nit: Please indent this line

src/fetch/fetchWithRetries.js
zpao: Also, let's put `requestsAttempted` onto its own line

emmenko:
@zpao done the changes according to your comments, I've squashed the commits then.

src/fetch/fetchWithRetries.js
```diff
@@ -86,7 +86,13 @@ function fetchWithRetries(
             retryRequest();
           } else {
             // Request was not successful, giving up.
-            reject(response);
+            var error: any = new Error(sprintf(
+              'fetchWithRetries(): Still no successful response after ' +
+              '%s retries, giving up',
```
yungsters: One last nit. Period at the end of this? :)

emmenko:
@yungsters done :)

yungsters:
Looks good, thanks!

PR #54, opened 2015-09-20
Make tests pass on none facebook stack

clentfort:
Running `npm test` would fail with an error "Cannot find module './core-js/es6'
from '/fbjs/scripts/jest/environment.js'" Add `core-js` as a dev-dependency and
add it to the modules not being touched by the "rewrite modules"-script.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
   "license": "BSD-3-Clause",
   "devDependencies": {
     "babel": "^5.4.7",
+    "core-js": "^1.1.4",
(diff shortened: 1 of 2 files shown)


zpao:
`core-js` is already a dependency. Also, shouldn't be rewriting modules when testing (except for ones in explicitly in the map).

I'm curious how you have issue with this (fresh clone and install worked just fine here)…

clentfort:
After upgrading to node version 4 it works, with iojs 3.somewhat it did not.

PR #55, opened 2015-09-21
Add gulp header plugin

zertosh:
@zpao, I was (fully) porting React from grunt to gulp, and I figured that this makes sense as a gulp plugin  and that it should live here.

cc: @cpojer

### DIFF
--- a/scripts/gulp/header.js
+++ b/scripts/gulp/header.js
@@ -42,7 +42,7 @@ const HEADERS = {
 
 class HeaderStream extends stream.Transform {
   constructor(opts) {
-    super({objectMode: true, highWaterMark: 16});
+    super({objectMode: true});
     this._header = HEADERS[opts.type](opts.pkg, opts.version, opts.startYear);
   }
 
--- a/scripts/gulp/header.js
+++ b/scripts/gulp/header.js
@@ -9,72 +9,80 @@
 
 'use strict';
 
-var gutil = require('gulp-util');
-var through = require('through2');
+const stream = require('stream');
+const gutil = require('gulp-util');
 
-var PLUGIN_NAME = 'header';
-var YEAR = '2015';
+const PLUGIN_NAME = 'header';
 
-var HEADERS = {
-  license: function(pkg, version) {
-    return [
-      '/**',
-      ' * ' + pkg + ' v' + version,
-      ' *',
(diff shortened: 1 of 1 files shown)


scripts/gulp/header.js
```diff
@@ -0,0 +1,80 @@
...
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+'use strict';
+
+var gutil = require('gulp-util');
+var through = require('through2');
+
+var PLUGIN_NAME = 'header';
+var YEAR = '2015';
```
zpao: `(new Date()).getFullYear()` maybe?

scripts/gulp/header.js
```diff
@@ -0,0 +1,80 @@
...
+      ' * Copyright 2013-' + YEAR + ', Facebook, Inc.',
+      ' * All rights reserved.',
+      ' *',
+      ' * This source code is licensed under the BSD-style license found in the',
+      ' * LICENSE file in the root directory of this source tree. An additional grant',
+      ' * of patent rights can be found in the PATENTS file in the same directory.',
+      ' *',
+      ' */',
+      ''
+    ].join('\n');
```
zpao: We'll be using a version of node that supports template literals soon and this will look much better then…

scripts/gulp/header.js
zertosh: Is it safe to trust the system time? I also thought about using "present".

scripts/gulp/header.js
zertosh: But then you'll have to use weird indentation so you don't get leading whitespace

scripts/gulp/header.js
zpao: Looks like "present" should be ok. Let's make the start year an option though since not all of our projects should have `2013`.

scripts/gulp/header.js
zpao:
```js
return `
/**
 * ...
 */`.trim()
```

Gross all around, which is part of why I moved to reading the template from disk in React.

zertosh:
I gave it the full node-4.0 compatible ES6 treatment - no destructuring. And since we can count on modern streams and use `class`, I ditched `through2`. If this is good, then I'll squash my changes.

PR #57, opened 2015-09-22
Implement UserAgent fork using ua-parser-js

zpao:
The goal here is to approximate our internal implementation, which is
all done server-side. This will not be perfect but we'll patch things up
as we need to.

We opted for ua-parser-js since it matches up pretty well with the handful of UAs we threw at it, has a fairly small footprint, is popular seems to be well supported.

cc @salier

### DIFF
--- a/src/__forks__/UserAgentData.js
+++ b/src/__forks__/UserAgentData.js
@@ -1,0 +1,84 @@
+/**
+ * Copyright 2013-2015, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @providesModule UserAgentData
+ */
+
+/**
+ * Usage note:
+ * This module makes a best effort to export the same data we would internally.
+ * At Facebook we use a server-generated module that does the parsing and
+ * exports the data for the client to use. We can't rely on a server-side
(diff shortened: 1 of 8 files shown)


zpao:
Another thing we might be able to do is match up more closely with our code so `UserAgent` stays in line with our internal code and we just have a `UserAgentData` module that exports the data (which is how it works internally). I was trying to avoid actually parsing at `require` time though, but it might not actually matter…

zpao:
Alright, after working on this some more, we decided to just keep our internal UserAgent module the same and instead implement UserAgentData. The downside of this is that it will be calculated at require time, but that shouldn't be too bad.

spicyj:
crazytown!

src/__forks__/UserAgentData.js
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Usage note:
+ * This module makes a best effort to export the same data we would internally.
+ * At Facebook we use a server-generated module that does the parsing and
+ * exports the data for the client to use. We can't rely on a server-side
+ * implementation in open source so instead we make use of an open source
+ * library to do the heavy lifting and then make some adjustments as necessary.
+ * It's likely there will be some differences. Some we can smooth over.
+ * Others should are going to be harder.
```
wallclockbuilder: Others ~~should~~ are going to be harder.

zpao: Thanks. Fixed with 2ff5001b4d461f9c84ed8bbd88cf5f706b7f46aa

PR #59, opened 2015-09-23
Fix getActiveElement so that it does not throw error

kamilio:
when `document` does not exist

While using great library `react-unit` to test the React components I discovered that this function is the only part that needs `document` global otherwise it crashes (variable not defined). This is an attempt to make it work even without document.

I've added simple test to verify basic functionality, but haven't figure out how to test my scenario when `document` is not present.

Thanks a lot and looking forward for the feedback...

### DIFF
--- a/src/core/dom/getActiveElement.js
+++ b/src/core/dom/getActiveElement.js
@@ -14,9 +14,13 @@
  * Same as document.activeElement but wraps in a try-catch block. In IE it is
  * not safe to call document.activeElement if there is nothing focused.
  *
- * The activeElement will be null only if the document body is not yet defined.
+ * The activeElement will be null only if the document or document body is not yet defined.
  */
 function getActiveElement() /*?DOMElement*/ {
+  if (typeof document === 'undefined') {
+    return null;
+  }
+
   try {
     return document.activeElement || document.body;
   } catch (e) {
--- a/src/core/__tests__/dom/getActiveElement-test.js
+++ b/src/core/__tests__/dom/getActiveElement-test.js
@@ -1,0 +1,22 @@
+/**
+ * Copyright 2013-2015, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+'use strict';
+
+jest.dontMock('getActiveElement');
+
+var getActiveElement = require('getActiveElement');
+
+describe('getActiveElement', () => {
+  it('returns body when there is no activeElement', () => {
+    var element = getActiveElement();
+    expect(element.tagName).toEqual('BODY');
+  });
+});


kamilio:
I signed the CLA

src/core/dom/getActiveElement.js
```diff
@@ -14,13 +14,15 @@
  * Same as document.activeElement but wraps in a try-catch block. In IE it is
  * not safe to call document.activeElement if there is nothing focused.
  *
- * The activeElement will be null only if the document body is not yet defined.
+ * The activeElement will be null only if the document or document body is not yet defined.
  */
 function getActiveElement() /*?DOMElement*/ {
+  var localDocument = (typeof document === 'undefined') ? {} : document;
```
zpao: I think if we go with this I would ultimately just shorten it - we know that we're going to return `undefined` when `typeof document === 'undefined'` - no point assigning to an empty object and entering a try/catch.

zpao:
cc @salier, @yungsters  for a sanity check before going further

yungsters:
Seems reasonable, but agree with @zpao.

```
if (typeof document === 'undefined') {
  return null;
}
// ...
```

kamilio:
Hello, early return makes sense, thanks. Pushed new commit into my branch. Please review.

kamilio:
Should I squash the commits or is it fine?

yungsters:
Squashing is ideal. :)

Soon, we will be able to automatically squash. But for now, manually squash is desired. Thanks!

kamilio:
Squashed and forced push.

Tell me more about the automatic squash, is it desired to have each PR squashed into single commit? Or is it going to be on demand, wherever sensible?

yungsters:
I'm in the process of getting this project automatically synchronized with our internal repository at Facebook. This is how Relay and React Native currently work.

Once that happens, each merged PR will automatically be squashed into one commit.

kamilio:
Thanks a lot, @yungsters and what is the process of releasing React. When should I expect this PR be part of it?

zpao:
We'll push out a new fbjs version from here soon and then update the React dependency. 0.14 should be going out soon and I'll make sure it includes this.

jakepusateri:
This got reverted in https://github.com/facebook/fbjs/commit/03ce16ef6de5d06b0436b7caafa392e655954e30 . Was that intentional? Upon upgrade to React 0.14.5, I'm getting many errors that look like this while running unit tests.
```
     ReferenceError: document is not defined
      at getActiveElement (node_modules/react/node_modules/fbjs/lib/getActiveElement.js:25:12)
````

zpao:
No, I don't think that was intentional. @yungsters - it looks like we never pulled this commit internally after merging :(

yungsters:
Acckkk... I'll re-commit this.

conartist6:
Sorry, but I am planning to to pull this out. When we add type checking to it becomes clear that the vast majority of facebook's code would not be prepared for this value to be undefined. Also this is basically an alias for writing document.activeElement, which is not something that should be written when there is no document. React itself has a lot of code which will run when document is not defined, and checks should be done there by checking the `ExecutionEnvironment.canUseDOM` property when appropriate. I have looked through React in an  attempt to determine if any current usage of getActiveElement could possibly be called when document is undefined, but I don't see any.

If you can provide your a way that I can reproduce your issue I will ensure that it stays fixed, otherwise I will assume that in the last year React fixed the root cause of the issue.

kamilio:
Go ahead, should be fine :) Thanks for asking.

PR #61, opened 2015-09-25
Switch to isomorphic-fetch

### LINKED ISSUE #60: Better fetch module for server-side compat
https://github.com/facebook/relay/issues/136 has relevant discussion. The proposed module to use is https://www.npmjs.com/package/isomorphic-fetch

zpao:
Fixes #60 

This should be pretty safe. With browserify and webpack packaging, this will use whatwg-fetch (exactly the same code we had).

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,8 +60,8 @@
   },
   "dependencies": {
     "core-js": "^1.0.0",
-    "promise": "^7.0.3",
-    "whatwg-fetch": "^0.9.0"
+    "isomorphic-fetch": "^2.1.1",
+    "promise": "^7.0.3"
   },
   "devEngines": {
     "node": ">=3",
--- a/scripts/babel/default-options.js
+++ b/scripts/babel/default-options.js
@@ -33,6 +33,6 @@ module.exports = {
   _moduleMap: {
     'core-js/library/es6/map': 'core-js/library/es6/map',
     'promise': 'promise',
-    'whatwg-fetch': 'whatwg-fetch'
+    'isomorphic-fetch': 'isomorphic-fetch'
   }
 };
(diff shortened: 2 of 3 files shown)


zpao:
Hmm, this actually needs a little bit more… going to try to actually build relay first (are some issues atm)

zpao:
Ok, confirmed we're all good with https://github.com/facebook/relay/pull/393.

I'm not wild about the versioning here. I didn't think about changes like this to `fbjs` needing corresponding bumps in `fbjs-scripts`. Oh well, we'll bump minors for both here.

zpao:
cc @boourns - can you test this branch again with RN? I'm hoping by preempting the require we can still export something useful if it exists.

boourns:
I will try it tonight or early tomorrow.

boourns:
OK.  This PR does not work on react native currently.

After some debugging I've found that the problem is the `@providesModule fetch` line here:
https://github.com/zpao/fbjs/blob/isomorphic-fetch/src/__forks__/fetch.js#L9

I don't quite understand why, but with `@providesModule` in place, the rest of `./fetch.js` is never executed when `fetchWithRetries.js` requires `./fetch` (you can confirm this with a `console.log`).  And what's returned by `require('./fetch')` in `fetchWithRetries.js` is a module object that looks like it has the `fetch` method available as `fetch.fetch`:
image

With the `@providesModule` line removed, `./fetch.js` runs as expected, exports the `fetch` function, and this PR functions as desired on react native.

boourns:
@zpao did my feedback make sense?

denvned:
Is the problem described by @boourns still actual? Are there any other problems that prevent this from being merged?

jeswin:
+1 to getting this merged. Server-side Relay takes a lot of manual patching right now.

spicyj:
Happened in #95.

PR #63, opened 2015-09-30
Bring in some modules from React

zpao:
Wanted to move a couple things our of React. I just threw them in core like everybody else but they can be moved later (the important part is getting them out of React). There are certainly more but these came up in a couple discussions recently.

Note: I still need to do this internally and we'll have to do a diff to React to remove and update the fbjs dep. It would be safe to land this and not update React immediately (they'll be under completely different paths after compilation).

`memoizeStringOnly` is used by UserAgent (#57 commented out in the code there)

`joinClasses` is used by a bunch of components and surely non-React code and I'd rather not map them into React if we open source any of them. Not to mention that there is nothing React specific about it.

cc @yungsters @spicyj

### DIFF
--- a/src/core/__tests__/joinClasses-test.js
+++ b/src/core/__tests__/joinClasses-test.js
@@ -1,0 +1,50 @@
+/**
+ * Copyright 2013-2015, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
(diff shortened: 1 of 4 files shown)


spicyj:
Sure. React won't even use joinClasses once we delete cloneWithProps.

spicyj:
wow I don't even have push access on this repo?

PR #66, opened 2015-10-11
browserify friendliness

zertosh:
* https://github.com/facebook/react/commit/6fc53e043876f5b53b144847faaeb7ea4398f9c3#diff-c3a9ce50ac626f89fa72fdcb514bf9e9R110 works fine for building the dist `react.js`, but users who `require('react')` aren't getting fbjs envified. So someone using browserify is getting the `process` shim included, which is like 2kb and has an empty env – so `process.env.NODE_ENV !== 'production'` is always true for fbjs stuff. I took the liberty of using `loose-envify` because it's fast (I'm using it in [invariant](https://github.com/zertosh/invariant) already) – but feel free to remove it in favor of the actual envify.

<del>* Even though `flow/include` is in `.gitignore`, because `"files"` is used (instead of `.npmignore`), the published package includes it.</del>

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "LICENSE",
     "PATENTS",
     "README.md",
-    "flow/lib",
+    "flow/",
     "index.js",
     "lib/",
     "module-map.json"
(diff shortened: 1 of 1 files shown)


zertosh:
An envify vs loose-envify result comparer: https://gist.github.com/zertosh/75dbed34b833ad4ed9b1

package.json
```diff
@@ -30,7 +30,7 @@
     "LICENSE",
     "PATENTS",
     "README.md",
-    "flow/",
```
zpao: This was right - `flow/include` is supposed to get shipped to npm so it can get consumed by packages depending on fbjs and using flow (eg, Relay). https://github.com/facebook/relay/blob/master/src/.flowconfig#L8.

We're waiting on the flow team for a better solution but need this in the meantime.

zertosh: Fixed! Sorry, the dup files looked like a bug.

zpao:
Thanks!

PR #70, opened 2015-10-15
Enable `comma-dangle` as an eslint warning

yungsters:
Enabling `comma-dangle` to warn if there is a missing trailing comma on arrays or objects that span multiple lines, and warn if there is a trailing comma present on single line arrays or objects.

Trailing commas simplify adding and removing items to objects and arrays, since only the lines you are modifying must be touched.

> This commit message was shamelessly jacked from [the ESLint website](http://eslint.org/docs/rules/comma-dangle.html).


zpao:
Excellent. I'll do another sync out of our rules to catch any other changes we made internally too.

PR #72, opened 2015-10-19
clean up inline-requires transform

voideanvalue:
Just some cleaning up to match up what I did internally. Doesn't affect functionality.


zpao:
👍

PR #74, opened 2015-10-22
Update to jest 0.6.0

cpojer:
updated createCacheKeyFunction to use the correct config object for creating the hash; this was a breaking change in jest.

### DIFF
--- a/scripts/jest/createCacheKeyFunction.js
+++ b/scripts/jest/createCacheKeyFunction.js
@@ -7,34 +7,34 @@
  * of patent rights can be found in the PATENTS file in the same directory.
  */
 
-var crypto = require('crypto');
-var fs = require('fs');
-var path = require('path');
+ 'use strict';
+
+const crypto = require('crypto');
+const fs = require('fs');
+const path = require('path');
 
 function buildCacheKey(files, base) {
-  return files.reduce(function(src, fileName) {
-    return src + fs.readFileSync(fileName);
-  }, base);
+  return files.reduce(
+    (src, fileName) => src + fs.readFileSync(fileName),
+    base
+  );
 }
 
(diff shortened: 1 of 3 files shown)


scripts/jest/createCacheKeyFunction.js
```diff
@@ -7,34 +7,34 @@
...
-  }, base);
+  return files.reduce(
+    (src, fileName) => src + fs.readFileSync(fileName),
+    base
+  );
 }
 
-var transformRoot = path.join(__dirname, '..');
-var cacheKeyFiles = [
+const transformRoot = path.join(__dirname, '..');
```
zpao: What version of node did `const` start working? I think it's fine but want to make sure I know what goes in our changelog if anything.

cpojer: I think const worked on 0.10 but only with `"use strict";` on top. It doesn't matter, jest doesn't work with anything pre node 4 :)

scripts/jest/createCacheKeyFunction.js
```diff
@@ -7,34 +7,34 @@
...
-  var cacheKey = buildCacheKey(files, cacheKeyBase);
+const cacheKeyBase = buildCacheKey(cacheKeyFiles, '');
 
-  return function(src, file, options, excludes) {
-    return crypto.createHash('md5')
-      .update(cacheKey)
-      .update(JSON.stringify([src, file, options, excludes]))
-      .digest('hex');
-  };
+module.exports = files => {
```
zpao: Oh I lied when I said this could be a patch change. We'll bump minor. Using the arrow means iojs breaks (but jest 0.5 still worked there so somebody *could* be using it).

PR #75, opened 2015-11-06
Add babel 6 version of the inline-require transform

tadeuzagallo:
Add a new version of the transform that will be used in react-native when upgrading to babel 6. I didn't put the tests in `__tests__` because of the other transforms that are babel 5-only compatible.


zpao:
This should be fine to ship whenever. Thanks for keeping it independent for the time being, we can move it to overwrite the current one once all of our things are using babel 6.

cc @cpojer for a sanity check

cpojer:
cc @voideanvalue

vjeux:
lgtm

scripts/babel-6/inline-requires-test.js
```diff
@@ -0,0 +1,69 @@
...
+  it('should inline single usage', function() {
+    compare([
+      'var foo = require("foo");',
+      'foo.bar()'
+    ], [
+      'require("foo").bar();'
+    ]);
+  });
+
+  it('should inline requires that are not assigned', function() {
```
voideanvalue: Technically, these are just left unaffected.

voideanvalue:
Looks good :). Thanks for adding some tests!

scripts/babel-6/inline-requires-test.js
tadeuzagallo: Yea, this was a terrible test name, I just kept going with the `should inline...`, sorry :(

PR #77, opened 2015-11-09
Update package.json

tadeuzagallo:
New version that includes babel 6's version of the inline requires transform.


zpao:
This can just be 0.3.1.

PR #79, opened 2015-11-11
[babel] Add auto-importer transform

DmitrySoshnikov:
Adds auto-importer transform (Babel 5 and 6 versions). We auto-import `Promise` module, and also `regeneratorRuntime`.

### DIFF
--- a/scripts/babel/default-options.js
+++ b/scripts/babel/default-options.js
@@ -10,8 +10,19 @@
 'use strict';
 
 var babelPluginModules = require('./rewrite-modules');
+var babelPluginAutoImporter = require('./auto-importer');
 var inlineRequires = require('./inline-requires');
-var plugins = [babelPluginModules];
+
+var plugins = [
+  {
+    position: 'after',
+    transformer: babelPluginAutoImporter,
+  },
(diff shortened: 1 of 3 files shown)


DmitrySoshnikov:
CC @zpao, @yungsters 

re: `regeneratorRuntime` - not sure we need it in the fbjs version of this transform.

DmitrySoshnikov:
Per discussion with @zpao, removed `regeneratorRuntime`, and added rewriter after the auto-importer.

scripts/babel/default-options.js
```diff
@@ -10,8 +10,19 @@
...
+var plugins = [
+  {
+    position: 'after',
+    transformer: babelPluginAutoImporter,
+  },
+  {
+    position: 'after',
+    transformer: babelPluginModules,
+  },
+];
```
zpao: Can we move this change into a second PR. That way we can ship the rest of this in a patch version

zpao: Ugh, I guess Relay does use this file so they would want it. Let's leave it and I'll ship it as a minor version.

PR #81, opened 2015-11-11
Sync out changes from FB

zpao:
~~There's a test failure to figure out still, but otherwise,~~ (re-disabled, which is what we were already doing) I wanted to get this out before we do some additional syncing.


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## [Unreleased]
 
+- `getActiveElement`: No longer handles a non-existant `document`
```
spicyj: existent

PR #83, opened 2015-11-24
Build .flow files next to .js files in lib

zpao:
I need to actually check that this is what we're expecting for the upcoming Flow release but I can fix that after we test.

cc @spicyj @gabelevi


spicyj:
I suppose this is breaking so it can be 0.6.

zpao:
Yea, this should be 0.6.

PR #84, opened 2015-11-24
Strip @providesModule directives

zpao:
This should save us from some future headaches with situations like RN where `@providesModule` copies of some of these exist.


spicyj:
Happened in #95.

skevy:
Question to @zpao and @spicyj - should we also be stripping `@providesModule` from the .flow.js files? Or does it not matter?

spicyj:
Yes, probably.

PR #85, opened 2015-12-02
import type * as Promise -> import type Promise

gabelevi:


zpao:
Thanks! I'll probably end up doing another sync from internal for everything else as well before the version.

gabelevi:
Great, thanks!

PR #87, opened 2015-12-04
Create rewrite-modules.js

AlexJuarez:
ported rewrite modules to babel 6

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,8 @@
     "scriptPreprocessor": "node_modules/fbjs-scripts/jest/preprocessor.js",
     "setupEnvScriptFile": "node_modules/fbjs-scripts/jest/environment.js",
     "testPathDirs": [
-      "<rootDir>/src"
+      "<rootDir>/src",
+      "<rootDir>/scripts"
     ],
     "unmockedModulePathPatterns": [
       "<rootDir>/node_modules/",
@@ -60,6 +61,7 @@
   },
   "dependencies": {
     "core-js": "^1.0.0",
+    "babel-core": "^6.3.17",
     "loose-envify": "^1.0.0",
     "promise": "^7.0.3",
     "ua-parser-js": "^0.7.9",
@@ -67,7 +69,7 @@
   },
   "devEngines": {
     "node": ">=3",
-    "npm": "2.x"
+    "npm": ">=2.x"
   },
   "browserify": {
     "transform": [
--- a/scripts/babel-6/rewrite-modules.js
+++ b/scripts/babel-6/rewrite-modules.js
@@ -1,0 +1,147 @@
+/**
+ * Copyright 2013-2015, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+'use strict';
+
+var p = require('path');
+
+/**
+ * plugin options have changed see https://babeljs.io/docs/plugins/#options
+ * for details.
+ */
+function getModuleMap(state) {
+  return state.opts._moduleMap || {};
+}
+
+function getModulePrefix(state) {
+  return state.opts._modulePrefix;
+}
+
+/**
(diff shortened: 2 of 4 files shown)


scripts/babel-6/rewrite-modules.js
```diff
@@ -0,0 +1,113 @@
...
+        );
+      }
+    }
+  }
+
+  /**
+   * Transforms either individual or chained calls to `jest.dontMock('Foo')`,
+   * `jest.mock('Foo')`, and `jest.genMockFromModule('Foo')`.
+   */
+  function transformJestCall(context, call) {
```
steveluscher: Can you test this with the following code, and see what it produces?

```
jest.mock('foo').dontMock('foo').genMockFromModule('foo');
```

I get:

```
jest.mock('./././foo').dontMock('././foo').genMockFromModule('./foo');
```

Not that you broke it, but I'm pretty sure this never worked.

scripts/babel-6/rewrite-modules.js
```diff
@@ -0,0 +1,113 @@
...
+    ) {
+      return undefined;
+    }
+    var moduleArg = call.arguments[0];
+    if (moduleArg && moduleArg.type === 'Literal') {
+      var module = mapModule(context, moduleArg.value);
+      if (module) {
+        return t.callExpression(
+          call.callee,
+          [t.literal(module)]
```
steveluscher: The Babel 6 version of this is `t.stringLiteral`.

scripts/babel-6/rewrite-modules.js
```diff
@@ -0,0 +1,113 @@
...
+      t.memberExpression(object, member.property),
+      args
+    );
+  }
+
+  return {
+    visitor: {
+      CallExpression: {
+        exit(path) {
+          return (
```
steveluscher: Visitor methods in Babel 6 plugins don't need return values anymore; just operate on the nodes and exit.

steveluscher:
Take a look at @fkling's awesome AST explorer. You can now switch it into Babel 6 mode; simply copy and paste your plugin in, and hack until it covers every case! http://felix-kling.de/esprima_ast_explorer/#/a0H3ntJzOW

AlexJuarez:
I noticed I made quite a few mistakes when I wrote some of the transition. I have a better version forth coming.

AlexJuarez:
Bump, whats needed to move forward?

steveluscher:
I'd love to see a few test cases that cover all the cases. I was able to copy and paste the latest implementation into the AST explorer and break it: http://astexplorer.net/#/VGbbBZzB7V/2

zpao:
Please include a test plan in your description above. While we don't have explicit tests for these, it would be good to know what exactly was done to ensure that the transform works.

AlexJuarez:
Ok, so I have reworked the code a little bit to use path.traverse. Additionally I have added tests for the new plugin to verify that it works.

steveluscher:
It works! Leaving to @zpao to merge.

steveluscher:
yh31dhh9do-4

scripts/babel-6/rewrite-modules.js
```diff
@@ -0,0 +1,147 @@
...
+  return {
+    visitor: {
+      CallExpression: {
+        exit(path, state) {
+          if (path.node.seen) {
+            return;
+          }
+          transformRequireCall(path, state);
+          transformJestCall(path, state);
+          path.node.seen = true;
```
steveluscher: Just for your interest… I found `shouldSkip` in the handbook; maybe it's designed to do the same thing? https://github.com/thejameskyle/babel-plugin-handbook#paths

AlexJuarez: I wonder if shouldSkip would skip all further transformations for that node, when I only want to skip the one I have defined.

package.json
```diff
@@ -60,6 +61,7 @@
   },
   "dependencies": {
     "core-js": "^1.0.0",
+    "babel-core": "^6.3.17",
```
zpao: I'm going to followup and move this - it's not a dep of fbjs and really shouldn't even be a devdep (should be a devdep of fbjs-scripts, but we aren't running tests there right now so it's in the grey area)

PR #88, opened 2015-12-04
Update module-map.js

AlexJuarez:
add a use relative path flag option to allow @providesModule to be used locally.


AlexJuarez:
still a work in progress, no need to review yet.

spicyj:
@AlexJuarez Thanks for the clarification. As a general practice, it'd be appreciated if you can skip sending PRs until they're ready for review (or clearly mark that they're not ready in the PR description when you post it), to avoid wasting others' time looking at a PR that isn't ready for review yet.

PR #89, opened 2015-12-08
Upgrade ESLint & config

zpao:
This updates ESLint to the latest release and takes advantage of the new ability to export the config object from a JS file. This is essentially what we were already doing internally so this makes it much easier to keep our configuration in sync and keep comments. That's the bulk of the changes here, done in multiple commits to make the set of changes as obvious as possible (updating the config in yaml first, then converting to js). The rest is integration of that and then updating our open-source-only files to fix lint issues.


spicyj:
:+1:

PR #90, opened 2015-12-09
Add package.json for fbjs-haste package

spicyj:


zpao:
ok. you aren't going to have the fbjs/flow/lib for flow if you just publish src. Does that matter?

spicyj:
No, fbjs-haste just publishes the untransformed source files. RN uses them like this right now.

zpao:
👍

PR #93, opened 2015-12-29
[babel6] Add dev-expression plugin

zpao:
This replicates the behavior of the babel 5 plugin.

I really don't like that I'm setting and checking seen state on nodes, and I feel like that shouldn't be required. If we don't do this then we visit the new nodes again which gets us into an infinite loop of creating new nodes only to visit again.


zpao:
@voideanvalue @DmitrySoshnikov - is there a better way to make this work?

DmitrySoshnikov:
Perhaps `path.skip()` is a better alternative, than node mutation with `seen` flag. Also, we have in-house `invariant` transform, that does at least the first part of your plugin -- don't calls the function, if the condition is `false`.

PR #94, opened 2015-12-31
Use `warn` instead of `error` for warnings

sean-clayton:
`console.warn` is both visually and semantically correct for displaying warnings than `console.error`. I recommend switching to it :+1:

### DIFF
--- a/src/__forks__/warning.js
+++ b/src/__forks__/warning.js
@@ -39,7 +39,7 @@ if (__DEV__) {
       var argIndex = 0;
       var message = 'Warning: ' + format.replace(/%s/g, () => args[argIndex++]);
       if (typeof console !== 'undefined') {
-        console.error(message);
+        console.warn(message);
       }
       try {
         // --- Welcome to debugging React ---


zpao:
Thanks but we're not going to take this. We intentionally changed to using console.error to improve the debugging experience, namely that you get native logging of stack traces.

sean-clayton:
@zpao Why not use [`console.trace()`](https://developer.mozilla.org/en-US/docs/Web/API/Console/trace) in conjunction?

spicyj:
console.trace would probably be okay, though the yellow/red background is something we also want to keep. We also generally only use warning() for "must-fix" warnings where you really should fix them, so I don't mind the harshness of console.error.

spicyj:
At Facebook internally we use an `alert()`-like modal dialog for all warnings to ensure they're seen.

MrOpposite:
I have to agree with @sean-clayton on this. The facebook way of seeing warnings as more severe and display them as errors is really taxing on people used with the notion that "Big red bad error means the execution in a task has halted and not completed" while a warning means "Something went wrong during execution, but it did not necessarily halt."

Using big red messages saying "THE PROGRAM HAS HALTED AT THIS LINE PANIC!" when all you really mean is "You have a rather severe problem on line X, I can work around it, but fix it." goes against what both the browsers themselves are outputting into the console (errors for failed HTTP-requests, warnings when stuff do not have the correct content-type header, but the correct content-type header can be deduced etc.)

Not conforming to this de-facto standard within the JS community means you're in a sense isolating you're possibly hiding important fatal errors among your warnings, making debugging harder than it usually is. When I'm trying to fix a fatal error, I want the warning to be a warning, and the fatal error to be a fatal error. Warnings and errors are not equal!

mellis481:
@zpao How does using `console.error` in a function called `warning()` and where the message being logged starts with "Warning:" not make you feel like you're doing something wrong?

PR #95, opened 2016-01-03
[RFC] RN 0.18 Compatibility changes

skevy:
This PR performs the necessary changes to `fbjs` to allow for better compatibility with React Native, post the https://github.com/facebook/react-native/commit/6a838a4201f99ae5e88b3684bb798d363815dd53 commit.

TLDR; this PR does the following:

1. Strips the `@providesModule` directives from the built, npm version of fbjs (courtesy of @zpao). This is no longer necessary after https://github.com/facebook/react-native/pull/5084 is merged.
2. Switch to `isomorphic-fetch`, in order to avoid causing weird `require` issues that happen when the `whatwg-fetch` module is required on React Native (courtesy of @zpao). Also fixes the issues discussed in https://github.com/facebook/fbjs/pull/61, which is where this commit is from.
3. Adds the React Native promise implementation as `Promise.native.js`. The `.native` suffix was added in https://github.com/facebook/react-native/commit/6a838a4201f99ae5e88b3684bb798d363815dd53, and allows the RN packager to require a different file for a given module.
4. Checks for existence of `ErrorUtils` before setting the `fbjs` version on `global`. This is needed as `ErrorUtils` on React Native is polyfilled with specific behavior.

**Note:  # 3 is most likely a temporary fix. After discussing with @spicyj, the desired result would be to eventually remove the packager complexity introduced in https://github.com/facebook/react-native/commit/6a838a4201f99ae5e88b3684bb798d363815dd53. The reason I'm going for this approach at the moment is that it's only one module that needs this shimming in `fbjs`, and I think the value it brings, given the fact that this PR, along with the corresponding PR in react-native, allows for Relay to work OOTB, outweighs its small of amount of technical debt. It's very possible you all may disagree, and I'm happy to change the approach if we come up with a better solution. :)**

### DIFF
--- a/scripts/gulp/strip-provides-module.js
+++ b/scripts/gulp/strip-provides-module.js
@@ -1,0 +1,36 @@
+/**
+ * Copyright 2013-2015, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+'use strict';
+
+var gutil = require('gulp-util');
+var through = require('through2');
+var PM_REGEXP = require('./shared/provides-module').regexp;
+
+module.exports = function(opts) {
+  function transform(file, enc, cb) {
+    if (file.isNull()) {
+      cb(null, file);
+      return;
+    }
+
+    if (file.isStream()) {
+      cb(new gutil.PluginError('module-map', 'Streaming not supported'));
+      return;
+    }
+
+    // Get the @providesModule piece out of the file and save that.
+    var contents = file.contents.toString().replace(PM_REGEXP, '');
+    file.contents = new Buffer(contents);
+    this.push(file);
+    cb();
+  }
+
+  return through.obj(transform);
+};
--- a/src/__forks__/Promise.native.js
+++ b/src/__forks__/Promise.native.js
@@ -1,0 +1,35 @@
+/**
+ *
+ * Copyright 2013-2014 Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ *
+ * This module wraps and augments the minimally ES6-compliant Promise
+ * implementation provided by the promise npm package.
+ *
+ */
+
+'use strict';
+
+var Promise = require('promise/setimmediate/es6-extensions');
+require('promise/setimmediate/done');
+
+/**
+ * Handle either fulfillment or rejection with the same callback.
+ */
+Promise.prototype.finally = function(onSettled) {
+  return this.then(onSettled, onSettled);
+};
+
+module.exports = Promise;
--- a/src/__forks__/Promise.native.js
+++ b/src/__forks__/Promise.native.js
@@ -1,19 +1,10 @@
 /**
  *
- * Copyright 2013-2014 Facebook, Inc.
+ * Copyright 2013-2016 Facebook, Inc.
  *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- * http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
  *
  * This module wraps and augments the minimally ES6-compliant Promise
  * implementation provided by the promise npm package.
--- a/src/__forks__/fetch.js
+++ b/src/__forks__/fetch.js
@@ -11,5 +11,11 @@
 
 'use strict';
 
-require('whatwg-fetch');
-module.exports = self.fetch.bind(self);
+// This hopefully supports the React Native case, which is already bringing along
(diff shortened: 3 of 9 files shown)


scripts/gulp/strip-provides-module.js
```diff
@@ -0,0 +1,38 @@
...
+      cb(null, file);
+      return;
+    }
+
+    if (file.isStream()) {
+      cb(new gutil.PluginError('module-map', 'Streaming not supported'));
+      return;
+    }
+
+    // Get the @providesModule piece of out the file and save that.
```
yungsters: Typo: out of

scripts/gulp/strip-provides-module.js
```diff
@@ -0,0 +1,38 @@
...
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+'use strict';
+
+var gutil = require('gulp-util');
+var through = require('through2');
+var PM_REGEXP = require('./shared/provides-module').regexp;
+
+// var PLUGIN_NAME = 'strip-provides-module';
```
yungsters: Delete?

src/__forks__/Promise.native.js
```diff
@@ -0,0 +1,36 @@
...
+require('promise/setimmediate/done');
+
+/**
+ * Handle either fulfillment or rejection with the same callback.
+ */
+Promise.prototype.finally = function(onSettled) {
+  return this.then(onSettled, onSettled);
+};
+
+
```
yungsters: Nit: Delete.

src/stubs/ErrorUtils.js
```diff
@@ -11,13 +11,17 @@
...
 
-var ErrorUtils = {
-  applyWithGuard(callback, context, args, onError, name) {
-    return callback.apply(context, args);
-  },
-  guard(callback, name) {
-    return callback;
-  },
-};
+if (global.ErrorUtils) {
```
yungsters: This is the only change here that I can imagine potentially causing havoc on Facebook, I'll double check and report back. Should be fine, though.

yungsters: Oh wait, this is in stubs. Never mind. :)

yungsters:
About to merge this, but wanted to double check. Have you tested this with the Relay examples from `master` to verify there aren't any regressions?

skevy:
@yungsters I tested with my test Relay application...but let me test with the examples. I don't presume there will be any regressions, but I'll report back here in a bit.

skevy:
Good news! It works! But, it brought to light test failures in Relay...fixed those in https://github.com/skevy/relay/commit/b8fa29a8259f41facae45d1128c9d307859087f9

src/stubs/ErrorUtils.js
zpao: This still comes from an internal file (in s_u) so we should triple check :P

src/__forks__/fetch.js
```diff
@@ -11,5 +11,11 @@
...
-require('whatwg-fetch');
-module.exports = self.fetch.bind(self);
+// This hopefully supports the React Native case, which is already bringing along
+// its own fetch polyfill. That should exist on `global`. If that doesn't exist
+// then we'll try to polyfill, which might not work correctly in all environments.
+if (global.fetch) {
+  module.exports = global.fetch.bind(global);
+} else {
+  module.exports = require('isomorphic-fetch');
+}
```
zpao: I never got around to revisiting the feedback about this in #61. If it works then great but I just wanted to make sure it was all good. Sounds like combined with the providesModule bits it might have fixed the issue.

src/__forks__/Promise.native.js
```diff
@@ -0,0 +1,35 @@
+/**
+ *
+ * Copyright 2013-2014 Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
```
zpao: Where is this file coming from that it is Apache licensed? RN and this are both BSD so we should make sure we have the appropriate header here.

zpao: Ah, I see https://github.com/facebook/react-native/blob/master/Libraries/Promise.js, which probably should be updated with BSD header. Let's do that here.

skevy: This was an oversight on my part. I copied and pasted from RN, and for some reason this had Apache license. Will fix.

src/__forks__/fetch.js
skevy: It works fine, both on web and native.

The RN packager actually still pulls in isomorphic fetch, unfortunately, because it doesn't know about conditional requires. But, there's no error like there was with whatwg-fetch.

src/__forks__/Promise.native.js
skevy: Fixed in https://github.com/skevy/fbjs/commit/7b58cbe2b6d6d99fa315a514f91480df1362d463

src/stubs/ErrorUtils.js
skevy: @zpao @yungsters Let me know if I need to update this...admittedly, I'm kind of working blind when it comes to how these changes affect the internal FB codebase.

spicyj:
Note: Before merging this, we need a way to shim the 'warning' module, which we do internally at Facebook so that we can log the frequency of warnings in our dev environments. (YellowBox should probably go through the same codepath.)

skevy:
@spicyj can you share how it's being shimmed now?

spicyj:
The module is blacklisted and we have another file with `@providesModule warning` that takes its place. I also have a `@providesModule fbjs/lib/warning` which just exports `require('warning')` which is necessary because that's how React requires it. It's hacky now.

skevy:
So, in RN then (post merging https://github.com/facebook/react-native/issues/5084), in the FB internal blacklist, couldn't you still do the same thing? Blacklist `fbjs/lib/warning` and then do a module with `@providesModule fbjs/lib/warning`? Or, blacklist `Libraries/vendor/fbjs/warning` and do a module with `@providesModule warning`

spicyj:
Hm, maybe. It doesn't work if anyone in fbjs does a relative require of `./warning`, but maybe that already doesn't work…

skevy:
Yah, that's what I was thinking. I'm not sure if this PR or https://github.com/facebook/react-native/pull/5084 actually change how you'd shim it -- besides making sure you guys are maintaining your own blacklist, which you already do anyway.

src/stubs/ErrorUtils.js
yungsters: @zpao Confirmed that this has not mapped from an internal file in quite a while, so we're good. :)

skevy:
Woo! Thanks @yungsters! You guys can close #61 and #84 now, as those were included here.

src/stubs/ErrorUtils.js
zpao: Thanks for checking! We should perhaps move this into https://github.com/facebook/fbjs/tree/master/src/__forks__ then (since stubs/ is actually a thing in s_u and that's where EventListener lives). Doesn't really matter for anybody except a couple people like me and you who actually mostly know this stuff :P.

boourns:
:tada:

PR #96, opened 2016-01-05
FBJS: Fix `getActiveElement` w/o `document`

yungsters:
Brings back #59, which was unintentionally reverted by 03ce16e.

### DIFF
--- a/src/core/dom/getActiveElement.js
+++ b/src/core/dom/getActiveElement.js
@@ -10,13 +10,19 @@
  * @typechecks
  */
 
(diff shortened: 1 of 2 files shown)


src/core/dom/getActiveElement.js
```diff
@@ -10,13 +10,19 @@
  * @typechecks
  */
 
+/* eslint-disable fb-www/typeof-undefined */
```
yungsters: Added this, though.

zpao: 👍

PR #97, opened 2016-01-05
Update CHANGELOG.md for 0.7.0

yungsters:
This is just a placeholder with a list of changes since 0.6.0. We can update / replace it when 0.7.0 is actually published.

cc @zpao

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 ## [Unreleased]
 
+## [0.7.0] - 2016-01-04
+
+### Added
+
+- `Promise` with `finally` method for React Native
+
+### Changed
+
+- `ErrorUtils`: Re-uses any global instance that already exists
+- `getActiveElement`: Handles a non-existent `document` (again)
+
 ## [0.6.0] - 2015-12-29
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Changed
 
 - `ErrorUtils`: Re-uses any global instance that already exists
+- `fetch`: Switched to `isomorphic-fetch` when a global implementation is missing
 - `getActiveElement`: Handles a non-existent `document` (again)
 
 ## [0.6.0] - 2015-12-29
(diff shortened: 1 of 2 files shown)


ide:
@zpao what's your thinking around the timeline for 0.7.0? @skevy's recent fbjs diffs fix some RN issues people are having and allow Relay to work with RN. So the earlier we get a new fbjs release the sooner RN can ship with Relay compat.

zpao:
I meant to followup on this sooner, sorry.

So I guess the question is, how will the cascading of fbjs versions work? I know we updated fbemitter explicitly to get all the RN deps using 0.6. But we aren't going to upgrade React for every fbjs release so that's not a maintainable approach.

ide:
Part of the fbjs and packager changes are to ignore some of the providesModule declarations so I think it'll be fine if there are multiple copies of fbjs. It might be necessary for react and fbemitter to upgrade to 0.7 one time (I don't think so though).

spicyj:
Yeah, starting with RN 0.19 it's fine to have more than one fbjs.

martinbigio:
@zpao, @spicyj we're blocked to land https://github.com/facebook/react-native/pull/5084 up until a new release of fbjs is done (unless we depend on a commit, which I guess is not ideal, right?). There's no rush, just wanted to get an idea when you were planning to release the next version :)

skevy:
Also, it's in the changelog, but https://github.com/facebook/fbjs/pull/100 should be pulled in as well.

ide:
Depending on the commit would actually be a good way to move forward but fbjs has some prepublishing scripts so we need to grab it from npm.

Tlvenn:
@zpao Any update on the timeline ? Sorry to sound impatient but I am kinda excited to see the day where Relay and RN just work together. Thanks in advance.

zpao:
Gonna take a closer look at things today.

zpao:
Pushed 8a84822137b52e8c187a965c0cf22a465f357503 which mostly overlaps with this (but doesn't give it the version number). I think I'll merge #105 and then call that 0.7.

skevy:
Thanks @zpao!

PR #98, opened 2016-01-06
Strip @providesModule from flow.js files

skevy:
Strips @providesModule from flow.js files, as we're already stripping them from the built .js files.

/cc @spicyj @zpao

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -68,6 +68,7 @@ gulp.task('lib', function() {
 gulp.task('flow', function() {
   return gulp
     .src(paths.lib.src)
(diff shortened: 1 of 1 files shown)


zpao:
Have you run `npm link` and confirmed that `flow check src` still works in Relay? I think this is probably right as Ben said (since Flow should only look at the .js.flow files if it already resolved a module) but just want to make sure.

zpao:
And I just tested. It looks like this actually breaks flow whereas `master` is fine.

skevy:
Yah. I just tested as well. Actually, that makes sense, given that flow is aware of providesModule.

PR #99, opened 2016-01-06
0.6.1 (NOT FOR MERGE)

zpao:
I'm going to actually make this a branch in the fb repo but putting it up as a PR just for a quick discussion.

The main issue here is that we need this in React ASAP (breaks some testing functionality). I could publish this as 0.6.1 and React from npm will pick it up no problem. But we need to publish a new browser version, so regardless, we need to ship a new version of React.

The other issue is that the version number is… perhaps not exactly semver. We probably should make this 0.7, but in order to avoid the cascading issue with upgrading React Native and other projects so everybody has the same version of fbjs, it's so much easier to just make it 0.6.1. It's sort of a bugfix since we didn't mean to revert the change…

cc @ide since we've talked about the semver issue. I think this is probably fine but wanted to get a second opinion before I went ahead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## [Unreleased]
 
+## [0.6.1] - 2016-01-06
+
+### Changed
+- `getActiveElement`: no longer throws in non-browser environment (again)
+
 ## [0.6.0] - 2015-12-29
 
 ### Changed
--- a/src/core/dom/__tests__/getActiveElement-test.js
+++ b/src/core/dom/__tests__/getActiveElement-test.js
@@ -1,0 +1,22 @@
+/**
+ * Copyright 2013-2015, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+'use strict';
+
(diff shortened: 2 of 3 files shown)


zpao:
Well, I guess this technically fits under the 0.x.y exemption…

> Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.

ide:
This looks fine to me in practice. I'm sure there's some hypothetical project setup that this breaks -- you'd need to relying on the fact that getActiveElement would throw an error -- but on the quite unlikely chance this breaks someone's project they can simply set package.json to use 0.6.0, which seems less costly than having to make all these other projects upgrade to ^0.7.0. So this looks very pragmatic.

Regarding the 0.x.y exemption -- I believe npm's ^ caret is a bit stricter than what the semver docs say. 0.x.y versions are treated as if `x` were the major version -- so `0.[x+1].y` is for breaking changes and `0.x.[y+1]` is for API additions and bug fixes.

zpao:
Thanks for weighing in!

> Regarding the 0.x.y exemption -- I believe npm's ^ caret is a bit stricter than what the semver docs say. 0.x.y versions are treated as if x were the major version -- so 0.[x+1].y is for breaking changes and 0.x.[y+1] is for API additions and bug fixes. 

Yea, that's how it works. Which is sane and I think mostly in line with how people actually use 0.x.y.

zpao:
Shipped.

PR #102, opened 2016-01-09
shallowEqual returns true when dealing with NaN

### LINKED ISSUE #101: shallowEqual() should handle NaN values correctly
shallowEqual() should return true when dealing with NaN.

```
shallowEqual(NaN, NaN);  // true
shallowEqual({a: NaN}, {a: NaN});  // true
```

Fixing this would also fix the confusing error in React, see [React Issue 5773](https://github.com/facebook/react/issues/5773).

rickbeerendonk:
Fixes #101

### DIFF
--- a/src/core/__tests__/shallowEqual-test.js
+++ b/src/core/__tests__/shallowEqual-test.js
@@ -29,8 +29,22 @@ describe('shallowEqual', () => {
   it('returns true if arguments are shallow equal', () => {
     expect(
       shallowEqual(
-        {a: 1, b: 2, c: 3},
-        {a: 1, b: 2, c: 3}
+        {a: 1, b: 2, c: NaN},
+        {a: 1, b: 2, c: NaN}
+      )
+    ).toBe(true);
+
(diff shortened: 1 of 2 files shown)


Dashed:
You should also `Object.is` here: https://github.com/rickbeerendonk/fbjs/blob/83abbb77a41ec5feb8873b5427fddc80966e720c/src/core/shallowEqual.js#L24-L26

rickbeerendonk:
@Dashed That is a different issue, but you are correct in that it should also be handled.

syranide:
I'm not sure how perf-critical this is, but it could be worth considering `A === B || Object.is(A, B)` if it is.

zpao:
Thanks for doing this however… I'm going to close out because we already have this change internally and will be part of the next sync out. Sorry it's not already here - I need to be careful *when* it gets synced out for compatibility reasons

rickbeerendonk:
@zpao Thank you for explaining, no worries.

PR #108, opened 2016-02-02
Adding rejection-tracking promise module to module map

skevy:
Not having this module listed in the default options breaks `Promise.native.js` on build, as it `requires('./promise/setimmediate/rejection-tracking')` rather than the proper module sans the `./`.

/cc @zpao @martinbigio


skevy:
Also, be sure to delete and re-npm install before you build :)

zpao:
Thanks!

skevy:
No problem!

PR #109, opened 2016-02-03
[gulp] New script that checks that packages are up to date

zpao:
This came out of a need where an issue with a project was fixed by upgrading dependencies but other team members were still seeing the issue since upgraded packages aren't caught in anyway.


zpao:
Here it is in action after I explicitly installed an outdated version of `del`:
screen shot 2016-02-03 at 12 16 46 pm

PR #110, opened 2016-02-04
Update URI stub to have most basic expected functionality: stringification

zpao:
Was working on a project using URI and hit the expectation that stringifying an instance would give a string uri back.

### DIFF
--- a/src/stubs/URI.js
+++ b/src/stubs/URI.js
@@ -7,12 +7,22 @@
  * of patent rights can be found in the PATENTS file in the same directory.
  *
  * @providesModule URI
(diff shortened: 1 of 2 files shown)


src/stubs/URI.js
```diff
@@ -7,12 +7,22 @@
...
  *
  * @providesModule URI
- * @typechecks
+ * @stub
  * @flow
  */
 
 'use strict';
 
-class URI {}
```
spicyj: lol nice stub

spicyj:
okey

PR #111, opened 2016-02-05
Move RN Promise rejection tracking code out of fbjs

skevy:
After discussion with @davidaurelio, this PR moves the Promise rejection tracking code out of fbjs (really shouldn't have been in there in the first place), and instead just makes fbjs export a React Native compliant version of Promise, rather than adding functionality.

The corresponding change in RN will require the release of 0.7.2 of fbjs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 ### Fixed
 - `URI`: correctly store reference to value in constructor and return it when stringifying
 
+### Removed
+- Backed out rejection tracking for React Native `Promise` implementation. That code now lives in React Native.
+
 ## [0.7.1] - 2016-02-02
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


zpao:
Did any version of RN depend on this functionality (even if it was only there for a couple days)? I'm not sure that I'm comfortable taking away functionality in a 0.x.y release.

skevy:
No version of RN ever depended on this. 0.7 (and 0.7.1 for that matter) were released for usage in https://github.com/facebook/react-native/issues/5084

skevy:
It's the unfortunate part about npm sometimes...we had to release those in order to properly make tests past :(

zpao:
We should work on a way to make it so we don't need to release versions before they're ready.

skevy:
@zpao I agree with that sentiment.

zpao:
Can you also squash into a single commit (or 2 if you want to change the changelog separately) - helps keep history a bit cleaner

skevy:
@zpao np -- sorry i'm used to everything going through phabricator first! :-p

PR #114, opened 2016-02-19
[scripts] Check that all lib files can be required

### LINKED ISSUE #68: Make build fail when requires can't be resolved
Right now we don't care. But when adding new modules, ensuring all the dependencies are met might be a bit tedious. 

A quick sanity check (just `require` each file in `lib/` after transforming) should probably catch this.

zpao:
Quick and dirty but it works. It turns out I missed a dependency while working on #113 so this will now prevent the build from completing successfully. It's written so that any of our other projects should be able to use it too.

Fixes #68

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: node_js
 node_js:
   - "4"
 env:
+  - CMD=build
   - CMD=test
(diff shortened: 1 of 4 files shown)


spicyj:
:+1:

PR #115, opened 2016-02-19
[babel] Rewrite more jest methods

zpao:
Also slipped in a related-ish change, to preserve all arguments when
rewriting the first. This was required to support the `jest.setMock`
use.

### DIFF
--- a/scripts/babel-6/rewrite-modules.js
+++ b/scripts/babel-6/rewrite-modules.js
@@ -36,6 +36,18 @@ function getArguments(path) {
   }
 }
 
+var jestMethods = [
+  'dontMock',
+  'genMockFromModule',
+  'mock',
+  'setMock',
+  'unmock',
+];
+
+function isJestProperty(t, property) {
+  return t.isIdentifier(property) && jestMethods.indexOf(property.name) !== -1;
+}
+
 module.exports = function(babel) {
 
   var t = babel.types;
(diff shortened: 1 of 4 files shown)


scripts/babel-6/rewrite-modules.js
```diff
@@ -76,26 +93,26 @@ module.exports = function(babel) {
...
+      isJestProperty(t, calleePath.node.property)
     ) {
-      var module = mapModule(state, args.value);
+      var module = mapModule(state, moduleArg.value);
       if (module) {
-        args = t.stringLiteral(module);
+        args[0] = t.stringLiteral(module);
       }
     }
     path.replaceWith(
```
spicyj: Can you do `path.get('arguments').get(0).replaceWith(t.stringLiteral(module))`? That probably doesn't work for some reason but it seems worth asking…

scripts/babel-6/rewrite-modules.js
```diff
@@ -36,6 +36,18 @@ function getArguments(path) {
...
+var jestMethods = [
+  'dontMock',
+  'genMockFromModule',
+  'mock',
+  'setMock',
+  'unmock',
+];
+
+function isJestProperty(t, property) {
+  return jestMethods.some((name) => t.isIdentifier(property, {name}));
```
spicyj:
```
t.isIdentifier(property) && jestMethods.includes(property.name)
```

seems simpler?

spicyj:
k

scripts/babel-6/rewrite-modules.js
zpao: `path.get('arguments')` is an array so no luck. Good idea though.

zpao: *but* `path.get('arguments')[0].replaceWith(t.stringLiteral(module))` does work (since that's a `path`). Might rework a few other lines to account for that…

scripts/babel-6/rewrite-modules.js
zpao: `includes` isn't in node yet so good ol' fashioned `indexOf !== -1` it is

Issue #118, opened 2016-02-25
reletavise requires in compiled .flow files


zpao:
There's an issue right now where the result is that we basically force all indirect consumers of fbjs (eg using Relay) to use the haste system in their package or ignore things specifically. @mroch said we should be able to make this work with our projects and others by rewriting the requires in our generated flow files so they are relative while keeping the `@providesModule` headers.

samwgoldman:
mine

zpao:
👍 we might need to upgrade Babel for this to work, at least the plan I had
was to just run through Babel with only the existing requires pouring we
use for js. But that's not possible right now to run through and only do
that with Babel 5 (at least in the 2 minutes I tried)

samwgoldman:
Thanks—pretty eager to get this fix out. Does a Babel upgrade have any blockers to landing, other than just getting it to work locally?

zpao:
I think the only thing I'd check is diff the JS files before and after to make sure we don't slip anything in there we don't expect. We have all the plugins ported already. And actually we're just about to do the upgrade for React in https://github.com/facebook/react/pull/6113

kittens:
Do we have anything else that uses Babel 5? I want to kill it with fire.

samwgoldman:
@kittens If you're on a tear and want to take this one, go for it. I think you'll be able to turn it around faster than me. :)

zpao:
Literally everything else besides React Native still uses 5.

kittens:
Like what? React Native, Relay, www, and now React all use Babel 6.

zpao:
Relay doesn't use Babel 6, nor does flux nor draft. Nothing using
fbjs-scripts did until today (all need the dev plugin you just wrote)

kittens:
Thanks for the list, I'll upgrade all of those next week..

On Saturday, 27 February 2016, Paul O’Shannessy <notifications@github.com>
wrote:

> Relay doesn't use Babel 6, nor does flux nor draft. Nothing using
> fbjs-scripts did until today (all need the dev plugin you just wrote)
>
> On Friday, February 26, 2016, Sebastian McKenzie <notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>>
> wrote:
>
> > Like what? React Native, Relay, www, and now React all use Babel 6.
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/facebook/fbjs/issues/118#issuecomment-189577574>.
> >
>
[...]

PR #120, opened 2016-02-27
Upgrade to Babel 6

samwgoldman:
Biggest changes:
1. Including all the various plugins that Babel 5 came with by default
2. Moving the module plugin config where it now belongs

(1) was mechanical, but I'm a Babel n00b, so please take a careful look at
that. I just copied what @kittens did for React +flow +object-rest-spread.

For (2) I changed the default-options module to export a function that
takes a bit of configuration. The module takes care of threading the
plugin options to the right place. This seemed better than deeply
mutating a static export.

Next step here is to apply the module rewrite plugin to .js.flow files,
which we should be able to do with Babel 6.

Test Plan: I compared the diff between the existing lib output and the
new lib output. The `"use strict";` pragma moved under the first
comment, but that seems fine. Some extra parens. Seems fine.


samwgoldman:
Ref #118

zpao:
Looks good. Here's the diff: https://gist.github.com/zpao/2c815bbe453f0a9a9281. The docblock moving is annoying but since we don't really need it (no `@providesModule` in the JS) that's fine. Maybe one day the reprinter will do better with comments (nudge nudge @kittens)

zpao:
Let's merge this and then do the flow thing next. Thanks Sam!

PR #121, opened 2016-02-28
Relativize requires in .flow.js files

### LINKED ISSUE #118: reletavise requires in compiled .flow files

samwgoldman:
Now that we are on Babel 6, it's easy to pull in only the transforms that we need, leaving the rest intact.

Fixes #118

Test Plan:
Took the diff of the current master's `lib` and this PR's. It's pretty noisy, but you can see that the local requires are relative and node_module requires remain absolute. https://gist.github.com/samwgoldman/028cf76abf1e072920c5

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -6,11 +6,13 @@ var del = require('del');
 var mergeStream = require('merge-stream');
 var runSequence = require('run-sequence');
 
-var babelPluginDEV = require('fbjs-scripts/babel-6/dev-expression');
 var babelDefaultOptions = require('fbjs-scripts/babel-6/default-options');
+var babelPluginDEV = require('fbjs-scripts/babel-6/dev-expression');
+var babelPluginModules = require('fbjs-scripts/babel-6/rewrite-modules');
 var gulpModuleMap = require('fbjs-scripts/gulp/module-map');
 var gulpStripProvidesModule = require('fbjs-scripts/gulp/strip-provides-module');
 var gulpCheckDependencies = require('fbjs-scripts/gulp/check-dependencies');
+var thirdPartyModuleMap = require('fbjs-scripts/third-party-module-map.json');
 
 var paths = {
   lib: {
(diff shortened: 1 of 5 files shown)


gulpfile.js
```diff
@@ -67,6 +69,16 @@ gulp.task('lib', function() {
 gulp.task('flow', function() {
   return gulp
     .src(paths.lib.src)
+    .pipe(gulpModuleMap(moduleMapOpts))
+    .pipe(gulpStripProvidesModule())
```
zpao: I'm pretty sure we still need the providesModule bit for the projects that are using Haste (eg Relay) - see #98. We should do the same test as I did there.

samwgoldman: Hmm, I don't think we can have it both ways, though. Haste wants non-relative requires, Node wants relative, right?

zpao: I think Haste would resolve to this .flow file with providesModeul and then the requires within would be relative and still point to other modules which have a .flow file as well. I think Haste would use the `./` at the beginning of the requires here to drop out of pM resolution one it gets into a file. I think.

The problem is that we run flow check on the untransformed files in other projects. So until all of them only do `require('fbjs/whatever')`, then we have to make both work.

samwgoldman: OK, I'll test it. :)

samwgoldman:
@zpao Good call, re `@providesModule`. You were right. `flow check src` passing now in relay.

package.json
```diff
@@ -21,6 +21,7 @@
     "babel-eslint": "^5.0.0",
     "babel-plugin-check-es2015-constants": "^6.5.0",
     "babel-plugin-syntax-trailing-function-commas": "^6.5.0",
+    "babel-plugin-syntax-object-rest-spread": "^6.3.13",
```
zpao: We need `babel-plugin-syntax-flow` as well since we explicitly use it (at least for npm2 - npm3 will likely make it available at the same level when the transform is installed). That's why travis is failing to do anything right now.

samwgoldman: Thanks for the heads up. I should pay more attention instead of just throwing up a PR and moving on to the next thing. Updated and passing now.

zpao:
Awesome, thanks Sam! Want to squash and we'll merge (no shipit bot here with the autoflattening)

samwgoldman:
Squashed and force pushed :)

PR #124, opened 2016-03-04
Protect `isEmpty` if `Symbol` is `undefined`

steveluscher:


steveluscher:
Can someone double check my logic here? I inverted the invariant logic. @yungsters @zpao

yungsters:
Looks good to me.

zpao:
Can you make this change internally too?

spicyj:
I think there's also a Babel transform that could fall back to `@@iterator` automatically for us? Might be misremembering.

yungsters:
Copy the newest version from internal before you merge this.

zpao:
Took @yungsters' diff - 3c3b9d00f283d63f939d75543bf396e8e5a9fd4e

steveluscher:
👍

PR #125, opened 2016-03-04
Create a fork of `Map` just for React Native.

steveluscher:
This should let us eliminate `Map` from the React Native packager's blacklist.

### DIFF
--- a/src/__forks__/Map.native.js
+++ b/src/__forks__/Map.native.js
@@ -1,0 +1,10 @@
+/**
+ * Copyright (c) 2013-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+module.exports = require('Map');


steveluscher:
[rfc] What I expect this to do is to create a file called `lib/Map.native.js` but I couldn't get gulp to build me a `lib/` directory to test it. Thoughts, @zpao?

src/__forks__/Map.native.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * Copyright (c) 2013-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
```
steveluscher: Eliminated `@providesModule` so that we can kill this: https://github.com/facebook/react-native/blob/master/packager/blacklist.js#L21

zpao:
Based on the discussion - this won't actually be needed right? (I don't think this was going to do what you wanted anyway - that `require('Map')` would have been rewritten to `require('./Map')`.

steveluscher:
Indeed! Looks like this won't be needed after all. Simply removing it from the packager blacklist was enough.

PR #129, opened 2016-03-11
Bump devEngine requirement to node >=4

### LINKED ISSUE #128: Increase Requirement to Node v4
What is holding us back from increasing the node version requirement to v4? This will allow us to use a lot of niceties like arrow functions, concise object notation, `Object.assign`, etc.

yungsters:
Closes #128.

After this, we can start going crazy with all the ES2015 features.

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "ua-parser-js": "^0.7.9"
   },
   "devEngines": {
-    "node": ">=3",
+    "node": ">=4",
     "npm": ">=2.x"
   },
   "browserify": {


zpao:
👍

*some* of the es2015 features https://nodejs.org/en/docs/es6/

yungsters:
Do you want to merge this before — or after — you publish the next version?

zpao:
Now is fine, it's doesn't affect end users just us.

kittens:
Is `devEngines` actually a thing? Can't find docs for it anywhere.

yungsters:
@kittens It's only a thing if you use https://github.com/facebook/fbjs/blob/master/scripts/node/check-dev-engines.js#L29, heh.

zpao:
Yea, I made it up because the whole node 0.10, 0.12, iojs, 4, 5 mess combined with needing npm2 for some of our flowconfigs meant we had no guarantees anybody was using the right thing and would just get seemingly random failures.

PR #130, opened 2016-03-12
Fixes broken scrollTop position

AlastairTaft:
Else the `scrollable.scrollHeight` will always be the same as the `viewport.clientHeight` which causes the max scroll Y position to always be `0` (the top of the page).

### DIFF
--- a/src/core/dom/getScrollPosition.js
+++ b/src/core/dom/getScrollPosition.js
@@ -36,7 +36,7 @@ function getScrollPosition(scrollable) {
 
   const viewport =
     scrollable === documentScrollElement ?
-      document.documentElement :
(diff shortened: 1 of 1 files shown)


AlastairTaft:
Causes [this issue](https://github.com/facebook/draft-js/issues/14).

AlastairTaft:
I don't know about the other use cases but for this case the code would work if it was simplified down to the below, as the `scrollable` is the `document.body`

```javascript
  // viewport var removed
  // ...
  const xMax = scrollable.scrollWidth - scollable.clientWidth;
  const yMax = scrollable.scrollHeight - scrollable.clientHeight;
```

PR #127, opened 2016-03-11
Create `babel-preset-fbjs` Package

yungsters:
Sorry for the huge update, but this cleans up a lot of stuff.

First of all, it pulls out the babel configuration into a new package named `babel-preset-fbjs`. The preset is completely compatible with all the new Babel 6 goodness.

I also removed any extraneous package dependencies and upgraded most of the rest.

Finally, I cleaned up all of the lint errors and warnings, updated some copyright headers.

**Test Plan:**
```
(cd babel-preset; npm install; npm link)
(cd scripts; npm link babel-preset-fbjs; npm install; npm link)
(npm link babel-preset-fbjs; npm link fbjs-scripts; npm install)
npm test
npm run test-plugins
npm run typecheck
npm run lint
```

PR #135, opened 2016-04-13
Babel preset & cleanup

zpao:
This revives #127. @yungsters is out for a little while so not going to put this on him.

- [x] Make it so that `.js.flow` files can also be generated with the preset so project doesn't have to declare any plugins explicitly
- [x] Cleanup some of the dependencies (some were added or moved in the split process, needs a pass)
- [x] Make sure tests are working (`Scroll` is failing)
- [x] Remove the remaining Babel 5 stuff from `fbjs-scripts` (will likely just ship 0.7 after this is done with this being the only change).


zpao:
Ok, rebased everything from the beginning so we shouldn't be introducing any unnecessary deps and un-rearranged the package.jsons for easier diffing (I know, npm did it but it was excruciating to rebase so hopefully I didn't screw that up).

Last but not least, here's the diff of `lib/`: https://gist.github.com/zpao/5e8d8b2badf781d6c638b42bc56fd862. As you can see the only difference is the .js.flow files for modules using Promise, they are now getting the autorequire transform running. Should probably change that back just to keep fbjs at status quo (it's a 1 line fix in the gulpfile)

zpao:
Alright, calling this done. Will merge & ship `babel-preset-fbjs@1.0.0` and `fbjs-scripts@0.7.0` in a few minutes.

cc @kassens

zertosh:
Is pulling out `babel-plugin-syntax-flow` too much of a breaking change to do a release of `fbjs` that's semver compatible with `^0.8.0` (i.e. `react@15` compatible)?

Now that babel 6 is [compiled with babel 6](https://github.com/babel/babel/blob/master/CHANGELOG.md#680-2016-05-02), the babel family of packages is using `babel-runtime@6`. In Nuclide we have a top-level `babel-runtime@5` (we're stuck with `babel@5` for the near-future – because Atom). fbjs, and few other packages we use, have undedupable `babel-runtime@6`s that nearly doubled our `node_modules` :( I'm trying to trim it where I can.

zpao:
Are you talking about `fbjs`, `fbjs-scripts`, or `babel-preset-fbjs`? We aren't using `babel-runtime` in `fbjs` itself so presumably you're talking about the preset but you don't have to use that.

zertosh:
I'm talking about `fbjs`. You're forced to install `babel-runtime` via `react@15.0.2 -> fbjs@0.8.1 -> babel-plugin-syntax-flow@6.8.0`.

```
└─┬ react@15.0.2
  ├─┬ fbjs@0.8.1
  │ ├─┬ babel-plugin-syntax-flow@6.8.0
  │ │ └─┬ babel-runtime@6.6.1
  │ │   └── core-js@2.3.0
  │ ├── core-js@1.2.6
  ...
  ```

I know React doesn't need `babel-plugin-syntax-flow` or `babel-runtime`, that's why I'm humbly asking if dcc63b332bd81559efc95864a1f90609f1578aca can be published as `fbjs@0.8.2` (i.e. semver compat with React's dep on `fbjs@^0.8.0`). I don't know if that'll break any other fbjs consumers https://www.npmjs.com/browse/depended/fbjs.

zpao:
oooohhh I see. I was confused because that's distinct from this PR :) That should never have been a dependency, I missed it in #121. It's definitely unused and we can publish fbjs@0.8.2 to fix.

PR #139, opened 2016-04-21
Sync from FB

zpao:
Probably fine to ship this as 0.8.2. Biggest difference is that Flow will have a few more annotations to work from

cc @wincent


wincent:
lg2m, although poor Travis seems to hate you.

PR #147, opened 2016-05-11
Add some things to the babel preset

zpao:
Mostly the goal here is to get rid of people needing to use the react and es2015 presets in addition to this. Ideally that's not necessary.

So I added everything the React preset uses that wasn't already in use. There isn't much left from the es2015 preset but there are still a few things excluded, mostly things we were blacklisting anyway (regex changes) and typeof symbol.

I also brought over the object-assign transform we started using in React. I don't think it makes sense to leave it over there. I need to double check but I should probably take this chance to see if we can make it play nicely with the object-rest-spread transform, or we might need to fork that (it loads in the `_extends` helper which uses `Object.assign` but maybe due to ordering where I saw this, that `Object.assign` call wasn't transformed).

This would be a major version bump.


babel-preset/configure.js
```diff
@@ -70,6 +74,11 @@ module.exports = function(options) {
       require('babel-plugin-transform-es3-property-literals'),
       require('babel-plugin-transform-flow-strip-types'),
       require('babel-plugin-transform-object-rest-spread'),
+      require('babel-plugin-transform-react-display-name'),
+      require('babel-plugin-transform-react-jsx'),
+      require('babel-plugin-transform-react-jsx-source'),
```
zpao: I actually think we don't want this. It will embed the full file path which means we'd be shipping `var _jsxFileName = '/Users/poshannessy/FB/code/draft-js/src/component/contents/DraftEditorContents.react.js'` to npm, which is pretty clowny. We could potentially make it a dev-only or opt-in transform but I'm inclined to just drop it. If consumers want it they can specify it themselves.

zpao:
Going to merge & ship as an alpha or something so we can run it for real in Draft. Will ship as final once it's actually used.

PR #148, opened 2016-05-12
Relativize type import in .flow.js files

tsing:
Generate relative type imports in .js.flow files

This is the diff of the current master's lib and this PR's:
https://gist.github.com/tsing/afa63600f4aa41c729f3eb2017c862eb

### DIFF
--- a/babel-preset/plugins/__tests__/rewrite-modules-test.js
+++ b/babel-preset/plugins/__tests__/rewrite-modules-test.js
@@ -33,6 +33,20 @@ describe('rewrite-modules', function() {
         expect(result.code).toEqual('require(\'test/test\');');
       });
 
+      it('should replace the prefix on type imports', function() {
+        let result = babel.transform(
+          'import type Test from "test";',
+          {
+            plugins: [
+              'syntax-flow',
+              [rewriteModules, {map: {'test': 'test/test'}}]
+            ],
+          }
+        );
+
+        expect(result.code).toEqual('import type Test from "test/test";');
+      });
+
       it('should transform jest and requireActual methods', function() {
         const code = `function test() {
           'use strict';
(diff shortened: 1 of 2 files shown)


babel-preset/plugins/__tests__/rewrite-modules-test.js
```diff
@@ -33,6 +33,20 @@ describe('rewrite-modules', function() {
         expect(result.code).toEqual('require(\'test/test\');');
       });
 
+      it('should replace the prefix on type imports', function() {
+        let result = babel.transform(
+          'import type Test from \'test\';',
```
zpao: Can you use backticks so we can avoid escaping? (here and in the `expect`)

babel-preset/plugins/__tests__/rewrite-modules-test.js
```diff
@@ -33,6 +33,20 @@ describe('rewrite-modules', function() {
...
+          {
+            plugins: [
+              "syntax-flow",
```
zpao: And then single quotes here (lint warns about this but just a warning so doesn't show up in travis, sorry)

zpao:
This looks great, thanks! Just a couple comments and we can get this out there.

babel-preset/plugins/__tests__/rewrite-modules-test.js
tsing: Use backticks for single line string will cause warnings in eslint quotes rule: `Strings must use singlequote.`

I changed it to `'import type Test from "test"'` to avoid escaping. I saw it is used this way in `inline-requires-test.js` too.

babel-preset/plugins/__tests__/rewrite-modules-test.js
tsing: Fixed.

zpao:
That quoting works for me :) Thanks!

PR #151, opened 2016-05-18
Add explicit type args

bhosmer:
Make sure all polymorphic type expressions have explicit type arguments, by adding explicit `any` arguments wherever implicit ones were being supplied by Flow. This is in preparation for an upcoming change which will require explicit arguments in all cases.


zpao:
Merged in www so going to take here now. Thanks for handling both!

PR #153, opened 2016-05-23
Remove "Failed Composite propType" checking from `warning`

### LINKED ISSUE #152: Remove "Failed Composite propType" checking from `warning`
facebook/react#6824 removed duplicate propType checking, therefore the blacklist at https://github.com/facebook/fbjs/blob/master/src/__forks__/warning.js#L34-L36 is no longer necessary. Should we remove it from here? @spicyj

keyanzhang:
Fixes #152. facebook/react#6824 removed duplicate propType checking, therefore the blacklist at https://github.com/facebook/fbjs/blob/master/src/__forks__/warning.js#L34-L36 is no longer necessary.

### DIFF
--- a/src/__forks__/warning.js
+++ b/src/__forks__/warning.js
@@ -31,10 +31,6 @@ if (__DEV__) {
       );
     }
 
-    if (format.indexOf('Failed Composite propType: ') === 0) {
-      return; // Ignore CompositeComponent proptype check.
-    }
-
     if (!condition) {
       var argIndex = 0;
       var message = 'Warning: ' + format.replace(/%s/g, () => args[argIndex++]);


spicyj:
Thanks, looks good. @zpao This should go in the next "major" (not patch) release – can we merge now or should we hold off?

zpao:
Let's hold off. I'm going to take #151 and probably call 0.8 done then unless something pressing comes up, but we could branch for that if we need.

amireh:
Hate to hijack, but is there any chance we can blacklist the contentEditable warning we get when using Draft.js?

It reads something like this: `A component is `contentEditable` and contains `children` managed by React. It is now your responsibility to guarantee that none of those nodes are unexpectedly modified or duplicated. This is probably not intentional.`

spicyj:
If you're on the latest version of React and Draft you shouldn't be seeing that warning.

amireh:
Hmm, we just barely made the 0.13 to 0.14 upgrade and I don't think the upgrade to 0.15 will happen anytime soon. Would you accept a PR for such a blacklist if I made one?

(I would've personally forked the package and avoided annoying upstream if it was a peerDependency of Draft/React but sadly it's not so that option isn't available to us.)

zpao:
No, we wouldn't accept a PR for that, sorry. We'd have to do another React 14 release to pick it up and it would involve branching here and it's ultimately not worth the effort and burden for a minuscule percentage of users. The path is to upgrade React (which *should* be pretty painless ultimately but I can understand not rushing into that).

spicyj:
The React 15 upgrade has very few breaking changes so I hope you can upgrade soon.

zpao:
Would it matter much if we removed this in 0.8.4? What's the impact on people who haven't upgrade to React 15.2?

spicyj:
Double prop types warnings for everything.

zpao:
We're not inadvertently doing the inverse and hiding the only warnings people would see?

spicyj:
I don't think so? Context: For years now, we ran prop types checks at element creation "Failed propType:" and at component mount "Failed Composite propType:". We filtered out the former at FB where we had a big legacy codebase, and filtered out the latter in open source with this check in fbjs. I recently cleaned up our internal callsites so we were able to switch to element-time validation everywhere, and filtering out "Failed Composite propType:" both at FB and in open source. Now no one was looking at those messages ever, so @keyanzhang removed the code that produced them – and we can remove the filter since it's not doing anything when used in React 15.2+.

zpao:
Oh gotcha. So current code is dead for 15.2 (well it's running and slowing down every warning call).

Let's just pull this into the next release then (soonish, will mostly just be #150) and then we can bump the dep in React for next version.

zpao:
Pushed in db26dc5.

PR #155, opened 2016-07-05
invariant: more strict check for format argument

yorkie:
The `format` argument actually must be a native `String` object, because it requires to own the `.replace` function, such that I improve the type check here might be better :-)

### DIFF
--- a/src/__forks__/invariant.js
+++ b/src/__forks__/invariant.js
@@ -31,7 +31,7 @@ function invariant(condition, format, a, b, c, d, e, f) {
 
   if (!condition) {
     var error;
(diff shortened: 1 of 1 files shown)


spicyj:
`format` should only ever be undefined or a string, so I don't agree that your change makes this code more correct. I think I even have a mild preference for the current way which will throw if you pass something else, alerting you to the error in your code.

yorkie:
> I think I even have a mild preference for the current way which will throw if you pass something else, alerting you to the error in your code.

So do you agree with the current way is not wild enough right? @spicyj

PR #159, opened 2016-07-11
Split packages

zpao:
We can't use Lerna (yet[*]) but ultimately the goal here is to set clearer boundaries between these packages.  It should also make it easier to further split up the packages and bring in more maintainers.

cc @yungsters - this is what I was talking about doing. It doesn't affect our other syncing plans much (I think).

[*]: This is because we currently do things like `require('fbjs-scripts/gulp/foo')`. Lerna makes package-level requires work but not deep ones.


yungsters:
Nice. It begins!

PR #171, opened 2016-08-17
Extract try/catch into a separate function to prevent deopt

gaearon:
Not a big deal but this can add up if a lot of elements are being created (such as the case with FixedDataTable).

Before:

screen shot 2016-08-17 at 17 33 22

After:

screen shot 2016-08-17 at 17 32 40


zpao:
Ok. Do you want this in a 0.8 to ship with React sooner? Master is 0.9 and has a few changes I'd rather not take in a React patch release (we should be able to do minor though)

gaearon:
It would be cool to ship together with my other changes but no big deal if cherry picking is too much pain.

PR #173, opened 2016-08-18
removing README.md

KrishMunot:
Moving the file from here to root


zpao:
We need that file, it's part of the package that ends up on npm. We should update it but definitely not remove it.

PR #188, opened 2016-09-28
Make DOM utilities work in nested browsing contexts

### LINKED ISSUE #156: Using `document` and `window` global variables won't cover all cases
Hi, 
I can see that we are replying to `window` and `document` variables, that won't be great for some edge cases.

For example on this instance:
https://github.com/facebook/fbjs/blob/master/src/__forks__/Style.js#L51
```js
var Style = {
  /**
   * Gets the style property for the supplied node. This will return either the
   * computed style, if available, or the declared style.
   *
   * @param {DOMNode} node
   * @param {string} name Style property name.
   * @return {?string} Style property value.
   */
  get: getStyleProperty,

  /**
   * Determines the nearest ancestor of a node that is scrollable.
   *
   * NOTE: This can be expensive if used repeatedly or on a node nested deeply.
   *
   * @param {?DOMNode} node Node from which to start searching.
   * @return {?DOMWindow|DOMElement} Scroll parent of the supplied node.
   */
  getScrollParent: function(node) {
    if (!node) {
      return null;
    }
    while (node && node !== document.body) {
      if (_isNodeScrollable(node, 'overflow') ||
          _isNodeScrollable(node, 'overflowY') ||
          _isNodeScrollable(node, 'overflowX')) {
        return node;
      }
      node = node.parentNode;
    }
    return window;
  },

};
```

then `Style.getScrollParent` won't work on the following example:
```js
var iframeEl = document.getElementById('some_iframe');
var scrollParentEl = Style.getScrollParent(iframeEl.contentWindow.document.getElementById('some_element'));
```

acusti:
Fixes #156. In the discussion in that issue, @spicyj wrote:

> If this is causing a problem in some Facebook project using fbjs, feel free to send a pull request.

I’ve seen a couple issues that I’m hoping can qualify, including:

facebook/react#427
facebook/draft-js#527

I can also say that I am working with a team on an editor app that would benefit greatly from having React, Draft.js, and fbjs all work seamlessly across iframes. When building a sophisticated editor of some kind, the use case of rendering from a single parent browsing context into a separate nested browsing context, treating it as a sandboxed rendering target that can publish UI events, is a really compelling one.

About the PR: We used eslint’s `no-undef` rule with a config that didn’t include the browser env to get a list of all files that use browser globals directly. We then went through each instance and adapted them to support use across nested browsing contexts where necessary.

Here’s a list of the files that use browser globals that we didn’t modify, along with an explanation of why. The only file we didn’t change where it could actually matter is `getViewportDimensions`, so I’ll start with that one:

`fbjs/src/core/dom/getViewportDimensions.js`

Uses global `document` and `window` extensively. Could be easily refactored to take an optional document argument, like `getDocumentScrollElement`, but I also couldn’t find it being used anywhere in fbjs or in React, so didn’t know if it was worth it or desired. Please let me know if I should implement that change.

`fbjs/src/__forks__/isEventSupported.js`
`fbjs/src/core/ExecutionEnvironment.js`
`fbjs/src/dom/translateDOMPositionXY.js`

Feature (or UA) detection against the global `window` / `document` objects

`fbjs/src/core/createNodesFromMarkup.js`
`fbjs/src/core/getMarkupWrap.js`
`fbjs/src/core/getVendorPrefixedName.js`

Uses `document.createElement()`, no nested browsing context issues

`fbjs/src/core/dom/getDocumentScrollElement.js`

Uses global `document` as fallback default (just like the changes we made in this PR for `getActiveElement.js`)

`fbjs/src/core/dom/getStyleProperty.js`

  Uses `window.getComputedStyle`, which seems to work fine across nested browsing contexts

There are also some polyfill / feature detections for fetch, requestAnimationFrame, Performance, ES2015 Map/Set that use globals, but I figured it wouldn’t be helpful to list those out.

### DIFF
--- a/packages/fbjs/src/__forks__/Style.js
+++ b/packages/fbjs/src/__forks__/Style.js
@@ -48,7 +48,8 @@ var Style = {
     if (!node) {
       return null;
     }
-    while (node && node !== document.body) {
+    var ownerDocument = node.ownerDocument;
+    while (node && node !== ownerDocument.body) {
       if (_isNodeScrollable(node, 'overflow') ||
           _isNodeScrollable(node, 'overflowY') ||
           _isNodeScrollable(node, 'overflowX')) {
@@ -56,7 +57,7 @@ var Style = {
       }
       node = node.parentNode;
     }
-    return window;
+    return ownerDocument.defaultView || ownerDocument.parentWindow;
   },
 
 };
--- a/packages/fbjs/src/core/dom/getElementRect.js
+++ b/packages/fbjs/src/core/dom/getElementRect.js
@@ -19,7 +19,7 @@ const containsNode = require('containsNode');
  * @return {object}
  */
 function getElementRect(elem) {
-  const docElem = document.documentElement;
+  const docElem = elem.ownerDocument.documentElement;
 
   // FF 2, Safari 3 and Opera 9.5- do not support getBoundingClientRect().
   // IE9- will throw if the element is not in the document.
--- a/packages/fbjs/src/core/dom/isNode.js
+++ b/packages/fbjs/src/core/dom/isNode.js
@@ -15,8 +15,10 @@
  * @return {boolean} Whether or not the object is a DOM node.
  */
 function isNode(object) {
+  var doc = object ? (object.ownerDocument || object) : document;
+  var defaultView = doc.defaultView || window;
   return !!(object && (
-    typeof Node === 'function' ? object instanceof Node :
+    typeof defaultView.Node === 'function' ? object instanceof defaultView.Node :
       typeof object === 'object' &&
       typeof object.nodeType === 'number' &&
       typeof object.nodeName === 'string'
--- a/packages/fbjs/src/__forks__/getUnboundedScrollPosition.js
+++ b/packages/fbjs/src/__forks__/getUnboundedScrollPosition.js
@@ -23,10 +23,10 @@
  * @return {object} Map with `x` and `y` keys.
  */
 function getUnboundedScrollPosition(scrollable) {
-  if (scrollable === window) {
+  if (scrollable.Window && scrollable instanceof scrollable.Window) {
     return {
-      x: window.pageXOffset || document.documentElement.scrollLeft,
-      y: window.pageYOffset || document.documentElement.scrollTop
+      x: scrollable.pageXOffset || scrollable.document.documentElement.scrollLeft,
+      y: scrollable.pageYOffset || scrollable.document.documentElement.scrollTop
     };
   }
   return {
--- a/packages/fbjs/src/core/dom/__tests__/getActiveElement-test.js
+++ b/packages/fbjs/src/core/dom/__tests__/getActiveElement-test.js
@@ -19,4 +19,17 @@ describe('getActiveElement', () => {
     var element = getActiveElement();
     expect(element.tagName).toEqual('BODY');
   });
+
+  it('uses optional document parameter when provided', () => {
+    var iframe = document.createElement('iframe');
+    document.body.appendChild(iframe);
+    var iframeDocument = iframe.contentDocument;
+    var element = getActiveElement(iframeDocument);
+    try {
+      expect(element.ownerDocument).toBe(iframeDocument);
+      expect(element.ownerDocument).not.toBe(document);
+    } finally {
+      document.body.removeChild(iframe);
+    }
+  });
 });
--- a/packages/fbjs/src/core/dom/getActiveElement.js
+++ b/packages/fbjs/src/core/dom/getActiveElement.js
@@ -18,15 +18,19 @@
  *
  * The activeElement will be null only if the document or document body is not
  * yet defined.
+ *
+ * @param {?DOMDocument} doc Defaults to current document.
+ * @return {?DOMElement}
  */
-function getActiveElement() /*?DOMElement*/ {
-  if (typeof document === 'undefined') {
+function getActiveElement(doc) /*?DOMElement*/ {
+  doc = doc || document;
+  if (typeof doc === 'undefined') {
     return null;
   }
   try {
-    return document.activeElement || document.body;
+    return doc.activeElement || doc.body;
   } catch (e) {
-    return document.body;
+    return doc.body;
   }
 }
 
--- a/packages/fbjs/src/core/dom/getScrollPosition.js
+++ b/packages/fbjs/src/core/dom/getScrollPosition.js
@@ -28,15 +28,15 @@ const getUnboundedScrollPosition = require('getUnboundedScrollPosition');
  * @return {object} Map with `x` and `y` keys.
  */
 function getScrollPosition(scrollable) {
-  const documentScrollElement = getDocumentScrollElement();
-  if (scrollable === window) {
+  const documentScrollElement = getDocumentScrollElement(scrollable.ownerDocument || scrollable.document);
+  if (scrollable.Window && scrollable instanceof scrollable.Window) {
     scrollable = documentScrollElement;
   }
   const scrollPosition = getUnboundedScrollPosition(scrollable);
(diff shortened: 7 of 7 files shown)


acusti:
I tried to add tests where there were any existing test files I could add to, which turned out to only include `packages/fbjs/src/core/dom/getActiveElement.js` and `packages/fbjs/src/core/dom/isNode.js`. The getActiveElement test case was easy, but the isNode one proved trickier. Here’s the test I wanted to add:

```js
  it('should accept nodes from different document contexts', function() {
    var iframe = document.createElement('iframe');
    document.body.appendChild(iframe);
    try {
      expect(isNode(iframe.contentDocument.body)).toBe(true);
    } finally {
      document.body.removeChild(iframe);
    }
  });
```

However, I realized when verifying it that the assertions always passed, even without our changes (i.e. with isNode relying on the global `Node` constructor). This led to the realization that jsdom has only a single DOM implementation (set of DOM classes) that it shares across any browsing contexts, which I [documented here](https://github.com/tmpvar/jsdom/issues/1453#issuecomment-249929785), so I don’t think there is currently a way to capture this behavior in a test case using jsdom, but instead would require an actual browser (and not Firefox, because it has surprising `instanceof` behavior across nested browsing contexts).

acusti:
I created a demo of a simple use case to illustrate the utility of this PR. I then duplicated the demo and have one using the current react and react-dom builds and one using built versions of those files made from the React branch I made that uses these changes with the changes in this branch npm linked into it.

**Without** the changes in this branch + the React PR: http://codepen.io/acusti/pen/kkBpWA?editors=0011
**With** the changes in this branch + the React PR: http://codepen.io/acusti/pen/rrpKYZ?editors=0011

acusti:
Here’s another version of the demo, but with a different implementation that illustrates the benefits of these changes more slightly more effectively:

**Without** the changes in this branch + the React PR: http://codepen.io/acusti/pen/EgQxJG?editors=0011
**With** the changes in this branch + the React PR: http://codepen.io/acusti/pen/qaxBby?editors=0011

packages/fbjs/src/__forks__/Style.js
```diff
@@ -48,15 +48,16 @@ var Style = {
...
+    while (node && node !== ownerDocument.body) {
       if (_isNodeScrollable(node, 'overflow') ||
           _isNodeScrollable(node, 'overflowY') ||
           _isNodeScrollable(node, 'overflowX')) {
         return node;
       }
       node = node.parentNode;
     }
-    return window;
+    return ownerDocument.defaultView;
```
spicyj: Can we do `ownerDocument.defaultView || ownerDocument.parentWindow` for IE8 compat?

packages/fbjs/src/core/dom/getElementRect.js
```diff
@@ -19,7 +19,7 @@ const containsNode = require('containsNode');
  * @return {object}
  */
 function getElementRect(elem) {
-  const docElem = document.documentElement;
+  const docElem = (elem.ownerDocument || document).documentElement;
```
spicyj: When is it appropriate to fall back to `document` here?

acusti: Never, good point. WIll update

acusti:
I updated with those changes, thanks @spicyj

packages/fbjs/src/core/dom/isNode.js
```diff
@@ -15,8 +15,10 @@
  * @return {boolean} Whether or not the object is a DOM node.
  */
 function isNode(object) {
+  var doc = object ? (object.ownerDocument || object) : document;
+  var defaultView = doc.defaultView || window;
```
spicyj: This will throw if doc is null, won't it?

acusti: I’m thinking `doc` can’t be null. It definitely passes the `isNode` tests.

If `object` is null, `doc` is `document`
If `object` is not null but `object.ownerDocument` is null, `doc` becomes `object`, which wasn’t null
If `object` is not null and `object.ownerDocument` is not null, it won’t throw

spicyj: Oh, my mistake. I missed the `|| object`. This seems good.

acusti:
@spicyj Do you have an opinion about `getViewportDimensions`? Should I modify the `get*` functions to take an optional `doc` argument?

spicyj:
Eh, let's leave it for now.

conartist6:
Look at the commit and original rationale for https://github.com/facebook/fbjs/pull/59. In getActiveElement.js. Either we should remove the check for undefined document or we should ensure that we do not reference document before we check to make sure its type is not undefined.

spicyj:
@conartist6 Do you mind sending a PR fixing this and a diff internally?

spicyj:
@conartist6 Good catch. Do you mind sending a PR fixing this and a diff internally?

conartist6:
Will do.

conartist6:
I think the fix is actually slightly more complicated because this function's return type in flow (which I'm adding and is why I'm looking) should be DOMElement (not DOMElement?). Calling in doesn't make sense if you know there's no document.

The original puller's issue should be fixed in react then, which can be trivially modified to guard the calls with `if (ExecutionEnvironment.canUseDOM) {` since we already know that react considers the result to be nullable.

spicyj:
I don't believe we call getActiveElement in React except when doing client-side rendering.

acusti:
I don’t believe React ever calls `getActiveElement` outside of SelectEventPlugin and ReactInputSelection which are definitely only used in client-side rendering.

I only kept the `if (typeof doc === 'undefined') {` check because it was the pre-existing behavior and I didn’t know where else that utility is used. Seems like removing that and changing the return value to `DOMElement` is a good idea and should work fine.

acusti:
@spicyj I have a PR (well, 2 of them, but they’re just alternative approaches to the same feature/fix) to React that depend on these changes: https://github.com/facebook/react/pull/7936 and https://github.com/facebook/react/pull/7866

Don’t know when this PR will be released, and I’m not sure how to coordinate this all. There’s also an updated version of jsdom that is now being required in master at jest, but there’s no new release available there either, and that would be nice to have in React for my PR because it would mean removing hacky jsdom workarounds from the tests I added. And then there’s the fact that those are all required for a PR I’m preparing to draft.js to make it independent of the global window / document objects (so it will work when rendered into an iframe), but I haven’t even opened that one yet.

Is there anything I can do to help with the coordination of all this? I’m particularly curious if there’s anything I can do to help with the React PR; my new tests are failing right now, but that’s because they depend on this PR, but I think it could still be worthwhile for someone to review it. Thanks for any info or recommendations you can provide! I truly and deeply appreciate the extraordinary effort and commitment you and the rest of the React ecosystem-related team at Facebook have been making for the open source stewardship of all these projects.

spicyj:
We're just really behind and I haven't had time to look. I don't think there's anything else we need from you right now. Sorry for the wait.

conartist6:
@acusti These changes are merged, what are you waiting on?

acusti:
@conartist6 Can’t be installed until a new release of `fbjs` gets published to npm

conartist6:
Ah, ok.

ExE-Boss:
With&nbsp;<https://github.com/jsdom/jsdom/pull/2691>&nbsp;merged and&nbsp;<https://github.com/jsdom/jsdom/issues/1453> finally&nbsp;fixed, the&nbsp;`isNode(…)`&nbsp;test can&nbsp;finally be&nbsp;implemented:
```js
it('should accept nodes from different document contexts', () => {
	const iframe = document.createElement('iframe');
	document.body.appendChild(iframe);
	try {
		expect(isNode(iframe.contentDocument.body)).toBe(true);
	} finally {
		document.body.removeChild(iframe);
	}
});
```

PR #194, opened 2016-11-07
Remove UnicodeUtilsExtra's dependence on Immutable

zertosh:
Use an object instead of a Immutable.Map, and remove Immutable. See D4141148.

### DIFF
--- a/packages/fbjs/src/unicode/UnicodeUtilsExtra.js
+++ b/packages/fbjs/src/unicode/UnicodeUtilsExtra.js
@@ -16,7 +16,6 @@
 
 'use strict';
 
-const Immutable = require('immutable');
 const UnicodeUtils = require('UnicodeUtils');
 
 /**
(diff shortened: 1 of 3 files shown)


packages/fbjs/src/unicode/UnicodeUtilsExtra.js
```diff
@@ -62,20 +61,22 @@ function getCodePointsFormatted(str) {
   return codePoints.map(formatCodePoint);
 }
 
-// We use this funky constructor instead of the object
+// We use this funky function instead of the object
 // constructor because JS object keys are always coerced
 // to string, but we want a map of int -> string pairs.
```
zpao: Based on the comment, seems like coercion is the reason, though it shouldn't really matter AFAIK.

zertosh:
@zpao that test failure looks unrelated.

zpao:
Yea, unrelated (but strange, I'll have to look into that)

PR #196, opened 2016-11-08
Change console.error to console.warn in warning function

MrOpposite:
I find that error logging is jarring when what's actually displayed is a warning. Really small change to use the console.warn instead of console.error to generate appropriate styling in the debug console.


zpao:
As discussed in #94, we're going to leave this as is. Thanks for taking the time though.

PR #199, opened 2016-11-29
[eslint-config-fbjs] New config based on Nuclide as "ideal" config

zpao:
Not sure this is quite in the right end state, but it's *most* of the way there. [Nuclide's config](https://github.com/dddfacebook/nuclide/blob/master/.eslintrc.js) has no errors but I think that's wrong so I took some of the settings from our current config. I also enabled a couple other things because I could. Raw diff from current Nuclide config: https://gist.github.com/zpao/ef06fe0725977663e6f9be1bba7f5635

I tried running it over React and it's still super noisy at this point. Many of these are things that we probably should do eventually but it wouldn't be super impactful at the moment. The following (in the current config now extending the new one here) makes it so we can focus on the real issues:

```js
    'arrow-parens': OFF,
    'camelcase': OFF,
    'no-console': OFF,
    'no-multiple-empty-lines': OFF,
    'no-param-reassign': OFF,
    'no-useless-escape': OFF,
    'no-var': OFF,
    'no-void': OFF,
    'object-curly-spacing': OFF,
    'object-shorthand': OFF,
    'prefer-arrow-callback': OFF,
    'prefer-const': OFF,
    'prefer-rest-params': OFF,
    'quote-props': OFF,
    'spaced-comment': OFF,
    'strict': OFF,

    'flowtype/delimiter-dangle': OFF,
    'flowtype/no-weak-types': OFF,
    'flowtype/semi': OFF,
    'flowtype/space-after-type-colon': OFF,
    'flowtype/space-before-type-colon': OFF,
    'flowtype/union-intersection-spacing': OFF,
    'prefer-object-spread/prefer-object-spread': OFF,
    'react/jsx-closing-bracket-location': OFF,
    'react/jsx-curly-spacing': OFF,
    'react/jsx-indent': OFF,
    'react/jsx-indent': OFF,
    'react/jsx-indent-props': OFF,
    'react/jsx-key': OFF,
    'react/jsx-space-before-closing': OFF,
    'react/jsx-tag-spacing': OFF,
    'react/no-deprecated': OFF,
    'react/no-direct-mutation-state': OFF,
    'react/no-unknown-property': OFF,
    'react/prop-types': OFF,
```

The result: https://gist.github.com/zpao/44d135cf9f6f797f013dc0366c9b763e. `no-return-assign` is noisy but seems to be often the result of assignments in arrow functions. The rest are easier fixes. Then a second pass would be to re-enable as many of those disabled ones as possible. Some are intentional to test deprecated React patterns.

cc @peterhal @zertosh @nmote @spicyj


zertosh:
> Nuclide's config has no errors but I think that's wrong

What I like about this is that only _syntax errors_ are shown as _errors_ during development, so as you're working, you can safely ignore all warnings (but not errors). At land time, we use `--max-warnings=0`, so that effectively all warnings become errors.

zpao:
And in your honor @zertosh, I added a config that downgrades everything to a warning so you can keep living your blissful lie 😛 - `fbjs/opensource/warning` (just like I have one that upgrades everything to an error for the current config: `fbjs/strict`)

zpao:
cc @cpojer @dmitriiabramov. I know you guys have your own config in jest right now separate from fbjs's current config but this could potentially be a good replacement. Nuclide will likely switch over but since you're all really active right now as well and have a lot of code, it would be amazing if you could give it a shot and give some feedback. Keep in mind that it's meant to be and *ideal* config (if we were to start with a blank canvas) so it might feel churny. Look beyond the churn initially and let me know if the end state is still not good.

zpao:
Note: I'm going to merge this now and ship as 2.0.0-alpha.1 for initial use. Feedback can go in #204.

cpojer:
Happy to deprecate ours if you send us a PR and keep us as strict as we are now!

dmitriiabramov:
i can actually give it a try this week

zertosh:
@zpao I've made updates to Nuclide's config since you first copied it :(

PR #200, opened 2016-12-02
[signedsource] add signedsource

kassens:
Moves this module out of React Native for easier reuse.


kassens:
cc @cpojer @zpao for feedback on doing this right.

Daniel15:
@yungsters has an internal diff to modernize the code a bit, you should probably pull that version in rather than this older version :)

kassens:
Updated.

PR #203, opened 2016-12-06
Upgrade to flow-bin@0.36.0

kassens:


packages/fbjs/src/.flowconfig
```diff
@@ -2,9 +2,10 @@
 .*/__tests__/.*
 
 [include]
+../node_modules
```
kassens: This is needed when node_modules are installed.

PR #218, opened 2017-02-01
add mouse event for ReactWheelHandler handleScrollX / handleScrollY

weaseliam:
The event.preventDefault() in onWheel breaks the browser zoom in / out while holding the ctrl key and scrolling using the mouse wheel.
A fix for this can be done if we return false from handleScrollX / handleScrollY.
In order to return false we need to see that ctrl is pressed and for this we need the event.

In my opinion, I find that passing the event is useful and also this is a non breaking change since it is passed as the last argument in the list.

Please see https://github.com/schrodinger/fixed-data-table-2/issues/113 for the actual use case.


weaseliam:
Can someone please look at this PR ?
Thanks.

weaseliam:
Bump ?

Issue #217, opened 2017-01-30
Release 0.9.0 (non-alpha version)


dmnd:
Relay is currently depending on 0.8.1 because 0.9.0 is still in alpha (and has been for 7 months).

This is a problem for Relay because it ends up depending on core-js twice (see https://github.com/facebook/relay/issues/1421).

Can you please publish the current alpha as a real release?

zpao:
React 15.x probably won't upgrade past fbjs 0.8.x so you won't see the benefits you're looking for in that other issue. I'll have to go back and see what has changed but I'm pretty sure there was a big difference with shallowCompare that would require React doing a major bump.

As to @wincent's question about why there's even an alpha here… I think that was to get some code into the hands of Draft, and then I haven't done a sync in a months.

We can do a sync here and land a couple other things (#165 should probably ship), but like I said, you likely won't see the benefits you want yet. Maybe we could update core-js in the 0.8 branch.

dmnd:
Releasing a non-alpha would still set things up to work nicely once React 16 ships, assuming they also bump to 0.9.0. 16 is already a major change so probably they'd be ok with updating fbjs too?

> Maybe we could update core-js in the 0.8 branch.

That'd be nice in the meantime if you're up for it.

dmnd:
Closing since releases happened. Thanks!

PR #222, opened 2017-03-16
Make DOM utilities work in nested browsing contexts

acusti:
This is a PR of #188 (which was merged into master in October) into the fbjs-0.8.x branch. I’ve been trying to freshen up http://github.com/facebook/react/pull/7936 (wound up having to just create a new branch from master with http://github.com/facebook/react/pull/9184) and make sure it’s ready to go and was noticing that the tests I’d added were hanging, then I realized it’s because the changes in fbjs that the React PR depends on (#188) haven’t been released to the fbjs 0.8.x line.

I also see from #217 that no React 15.x release will upgrade to fbjs 0.9.x because it will entail a breaking change and therefore a new major release. So, can we just merge the updated functionality into fbjs-0.8.x and release a new version there?

Let me know any concerns. The API changes in this PR are 100% compatible with previous versions and the functionality changes were designed to not affect existing code in any way.

### DIFF
--- a/src/__forks__/Style.js
+++ b/src/__forks__/Style.js
@@ -48,7 +48,8 @@ var Style = {
     if (!node) {
       return null;
     }
-    while (node && node !== document.body) {
+    var ownerDocument = node.ownerDocument;
+    while (node && node !== ownerDocument.body) {
       if (_isNodeScrollable(node, 'overflow') ||
           _isNodeScrollable(node, 'overflowY') ||
           _isNodeScrollable(node, 'overflowX')) {
@@ -56,7 +57,7 @@ var Style = {
       }
       node = node.parentNode;
     }
-    return window;
+    return ownerDocument.defaultView || ownerDocument.parentWindow;
   },
 
 };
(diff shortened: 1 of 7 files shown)


zpao:
Thanks @acusti!

cc @spicyj @gaearon 

I don't think there should be any issue shipping this out in 0.8 but that will mean that React 15.* will get this automatically (which means further small differences between the npm version and the dist version).

Let me know if you have any concerns, otherwise I'll ship this soon.

spicyj:
ack 👍

zpao:
Just published v0.8.10. Thanks again!

ramesius:
Hey, we are seeing failures in our tests where we shallow render our components. Looks like it stems from this change, the removal of the following guard on document: `if (typeof document === 'undefined')`, is causing `ReferenceError: document is not defined`. Whatever is calling this function is not passing in `doc`, and since it is running in node there is no global `document`.

danez:
I have the same here in our tests.

acusti:
Yikes. The test / server-side rendering failures should be resolved by #223. @ramesius thanks for the prompt note and thorough explanation!

ramesius:
@acusti Thanks for the quick turn around on this one.

PR #223, opened 2017-03-21
Make getActiveElement work without document global

acusti:
🌶🔧 to address https://github.com/facebook/fbjs/pull/222#issuecomment-287978117 and https://github.com/facebook/fbjs/pull/222#issuecomment-288073788

Tests and server-side rendering are breaking when no global document exists (an oversight from the original PR). This seems like a clean way to solve it.

### DIFF
--- a/src/core/dom/getActiveElement.js
+++ b/src/core/dom/getActiveElement.js
@@ -23,10 +23,10 @@
  * @return {?DOMElement}
  */
 function getActiveElement(doc) /*?DOMElement*/ {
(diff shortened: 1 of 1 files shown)


acusti:
I will also make a PR against master with the same commit.

zpao:
@spicyj had merged #226 first and I cherry-picked that into the branch, so we're good. Thanks for the quick follow up though!

spicyj:
Ah I missed this one. Thanks for sending it!

PR #224, opened 2017-03-21
Make getActiveElement work without document global (for master)

acusti:
Dupe of #223, but for master.


zpao:
Due of my comment in #223 but for master :) Thanks again - I think the fix we took already is fine.

PR #232, opened 2017-04-14
Fix: use console.warn instead of console.error in DEV

probablyup:
Rationale:

1. The call sites I modified are actual warnings, so it's not appropriate to trigger what looks like a hard error to the user. It comes off as being relatively developer-hostile.

2. Some build systems monitor console.error and will fail their respective builds if anything looking like a hard error is logged. This is especially an issue now with all the prop-types and createClass deprecation messages that are not straightforward to find and fix.


probablyup:
@cpojer could you take a look at this when you get a chance? Thanks!

zpao:
I appreciate this but we won't be doing it. See #94 for some further discussion.

PR #240, opened 2017-05-17
Add Flow types to `invariant`

yungsters:


zpao:
Welp, looks like we didn't have a good flowconfig to actually pick up the `__DEV__` declaration and none of the files using `__DEV__` were flowified… I'll fix that.

PR #241, opened 2017-05-22
Type `keyMap`

davidaurelio:
Adds flow annotations to `keyMap`, so that the return type is an object with all keys of the input object, and `string` values.

### DIFF
--- a/packages/fbjs/src/key-mirror/keyMirror.js
+++ b/packages/fbjs/src/key-mirror/keyMirror.js
@@ -8,6 +8,7 @@
  *
  * @providesModule keyMirror
  * @typechecks static-only
+ * @flow
  */
 
 'use strict';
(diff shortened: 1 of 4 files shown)


zertosh:
👍 but who approves this?

davidaurelio:
Maybe @zpao?

packages/fbjs/src/key-mirror/keyMirror.js
```diff
@@ -32,7 +33,7 @@ var invariant = require('invariant');
  * @param {object} obj
  * @return {object}
  */
-var keyMirror = function(obj) {
+var keyMirror = function<T: {}>(obj: T): $ObjMapi<T, <K>(K) => K> {
```
zpao: This change is mostly in www already, difference is `Object` vs `{}` and `O` vs `T`. Can you pick one and update the other? https://fburl.com/otx6xumj

(I need to do a full re-sync at some point soon)

packages/fbjs/src/struct/PrefixIntervalTree.js
```diff
@@ -66,7 +66,7 @@ class PrefixIntervalTree {
   /**
    * Binary heap
    */
-  _heap: Array<number>;
+  _heap: Int32Array;
```
zpao: Can you update the www copy too? https://fburl.com/2q0k2y53

davidaurelio:
types synced internally

PR #242, opened 2017-05-24
Follow a breaking change in npm@4

watilde:
To fix https://github.com/facebook/draft-js/issues/914, I updated `check-dependencies` to handle a breaking change in `npm@4`. This is a quick-fix, but it's how npm handle the exitCode in the test: [npm/test/tap/outdated.js#L107-L108](https://github.com/npm/npm/blob/latest/test/tap/outdated.js#L107-L108).

Probably, we will have a better way eventually, but it may take more time.
+ https://github.com/npm/npm/pull/16703

### DIFF
--- a/packages/fbjs-scripts/gulp/check-dependencies.js
+++ b/packages/fbjs-scripts/gulp/check-dependencies.js
@@ -35,7 +35,10 @@ module.exports = function(opts) {
     });
 
     outdated.on('exit', function(code) {
-      if (code !== 0) {
+      // exitCode set to 1 in npm 4
+      if (code === 1) {
+        process.exitCode = 0
+      } else if (code !== 0) {
         cb(new gutil.PluginError(PLUGIN_NAME, 'npm broke'));
       }
 


packages/fbjs-scripts/gulp/check-dependencies.js
```diff
@@ -35,7 +35,10 @@ module.exports = function(opts) {
     });
 
     outdated.on('exit', function(code) {
-      if (code !== 0) {
+      // exitCode set to 1 in npm 4
+      if (code === 1) {
+        process.exitCode = 0
```
wwalser: Changing the exit code is unnecessary. 

I think a better approach is to move ```cb(new gutil.PluginError(PLUGIN_NAME, 'npm broke'));``` into the if statement that check ```failures.length ```. This way gulp only exits if packages are out of date. we ignore the exit code of npm outdated all together.

wwalser:
Changing the exit code is unnecessary.

I think a better approach is to move ```cb(new gutil.PluginError(PLUGIN_NAME, 'npm broke'));``` into the if statement that checks ```failures.length```. This way gulp only exits if packages are out of date. Ignore the exit code of ```npm outdated``` alltogether.

watilde:
I'm going to close this since https://github.com/facebook/fbjs/pull/244

PR #243, opened 2017-05-30
dummyNode in markup functions

tomtymczyszyn:
Moved _dummyNode_ constant inside _createNodesFromMarkup_ and _getMarkupWrap_ functions.

While working on React project and using React Test Utils with JSDOM for unit tests, it makes complication when the _dummyNode_ constant is set once on the initial run of app.
When JSDOM is used to create globals _document_ and _window_ variables during tests, it is possible to update the _ExecutionEnvironment.canUseDOM_ flag, however the _dummyNode_ is not updated with it so it is not consistent during tests.
Using it inside functions provide possibility to get updated _canUseDOM_ flag and create document element for proper unit test experience.

### DIFF
--- a/packages/fbjs/src/core/createNodesFromMarkup.js
+++ b/packages/fbjs/src/core/createNodesFromMarkup.js
@@ -19,12 +19,6 @@ const getMarkupWrap = require('getMarkupWrap');
 const invariant = require('invariant');
 
 /**
- * Dummy container used to render all markup.
- */
-const dummyNode =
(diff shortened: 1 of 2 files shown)


zpao:
I appreciate the PR however I'm going to have to close this.

Our recommendation (and React's recommendation) over the years is that you isolate these sorts of back and forth across some barrier so that you aren't changing global state. The `ExecutionEnvironement` module is not intended to change like that and so the code depending on it similarly doesn't expect changes.

To boot, moving that code into the function defeats its purpose - to reuse cached nodes.

PR #247, opened 2017-06-29
Wrapped exported setImmediate in a pass-trough function, so it doesnt…

Andarist:
… lose its global context - fixing "Invalid calling object" error in IE and Edge

It will fix issues in dependant packages such as https://github.com/facebook/draft-js/issues/1230

The issue was somehow introduced by webpack3 different module handling (aint sure about exact changes which caused it, didnt have time to investigate, but itd definitely related).

I wasnt sure about code-style, so if you let me know I can ammend it to use arrow function or smth.

### DIFF
--- a/packages/fbjs/src/__forks__/setImmediate.js
+++ b/packages/fbjs/src/__forks__/setImmediate.js
@@ -13,5 +13,6 @@
 
 // setimmediate adds setImmediate to the global. We want to make sure we export
 // the actual function.
-require('setimmediate')
-module.exports = global.setImmediate;
+require('setimmediate');
+module.exports = (...args) => global.setImmediate(...args);
+


Andarist:
Ive found out [the reason](https://github.com/webpack/webpack/issues/5111#issuecomment-309758504)

packages/fbjs/src/__forks__/setImmediate.js
```diff
@@ -14,4 +14,6 @@
 // setimmediate adds setImmediate to the global. We want to make sure we export
 // the actual function.
 require('setimmediate')
-module.exports = global.setImmediate;
+module.exports = function(fn, ms) {
+  global.setImmediate(fn, ms);
+};
```
zpao: This is fine without arrow functions but we're losing the ability to pass rest args. Let's use rest and call spread to keep them. This should work:

```js
module.exports = (...args) => global.setImmediate(...args);
```

Andarist: oh, always forgetting about possibility to pass more args, never used it, fixed!

Andarist:
@zpao the linked issue is about default webpack3 setting breaking `Array.from` imports, similar thing happens to `setImmediate` and `setTimeout`

Webpack3 introduces a new option to disable this bundle-size optimization (`strictThisContextOnImports`), but ofc its kinda a global setting for your bundle and that sucks. It's a reasonable default, cause otherwise most people wouldn't even know that there is something more to optimize there and exporting things depending on `this` as a module (!) is rather totally weird thing to do.

Unfortunately those few globals when exported by a module are somehow losing their appropriate javascript context, look at `setTimeout`:
**ok**
```js
var timeout = setTimeout
timeout(() => console.log('smth'), 1000)
```
**not ok**
```js
var obj = { timeout: setTimeout }
obj.timeout(() => console.log('smth'), 1000) // Uncaught TypeError: Illegal invocation
```

And because of the way how webpack3 wraps the imported module its causing this error now.

I agree that this is more a webpack3's issue, but the best fix for that would be for webpack3 to detect (with some kind of internal whitelist) what's getting imported and to wrap it appropriately automatically even for a default `strictThisContextOnImports: false` setting.

Can't know how hard this is to do for webpack and for sure we cannot know when this will get fixed there (if ever) and imho it would be a bad thing to do to just tell people to set this `strictThisContextOnImports` setting to `true` as a remedy for the problem.

I'll definitely keep an eye on the issue and we'll inform about the progress here and possibly send a reverting PR (if this one gets merged) if they fix it.

Andarist:
It seems it got fixed in webpack.

PR #254, opened 2017-07-19
Upgrade: Upgrade ESLint peerDep to support v4

BYK:
**Summary**

Trying to get ESLint v4 in yarnpkg/yarn#3962. Related: facebook/jest#4072

**Test plan**

CI.

### DIFF
--- a/packages/eslint-config-fbjs/package.json
+++ b/packages/eslint-config-fbjs/package.json
@@ -21,9 +21,9 @@
   },
   "peerDependencies": {
     "babel-eslint": "^7.1.1",
-    "eslint": "^3.11.1",
+    "eslint": ">=3.11.1",
     "eslint-plugin-babel": "^4.0.0",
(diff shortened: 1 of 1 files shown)


packages/eslint-config-fbjs/package.json
```diff
@@ -21,9 +21,9 @@
   },
   "peerDependencies": {
     "babel-eslint": "^7.1.1",
-    "eslint": "^3.11.1",
+    "eslint": ">=3.11.1",
```
zpao: Let's do `^3.11.1 || ^4.0.0` here

zpao: Actually, nevermind. I'm going to just do a wholesale update of this package and ship 2.0. What's in master is already on it's way to 2.0 (.0-alpha.1) so this would all need to be backported and that's no fun.

PR #257, opened 2017-07-24
add 'instrument' flag to createCacheKey fn

aaronabramov:
Jest uses an extra option to differentiate intstrumented cache files from non-instrumented (https://github.com/facebook/jest/blob/master/packages/jest-runtime/src/script_transformer.js#L70-L72)
and it seems like fbsource is failing to collect coverage without this option


mjesun:
@aaronabramov Thanks for taking care of this! :)

PR #258, opened 2017-07-24
Sync over argument arity fixes for stricter Flow updates

zpao:
Ignoring the eslint stuff, which I guess I forgot to push to master the other day (will fix in a bit)…

This should fix some of the issues that Relay is seeing, but it was pointed out that `then` should also be causing issues. Looking at https://github.com/facebook/relay/commit/68241f7c9126b6c44646ffb078df63c2e3810d43, it seems like those were probably just blanketed over with fixmes.


pranaygp:
`Deferred.js` should also have types for `Deferred#then` and `Deferred#catch`. Adding those :)

PR #259, opened 2017-07-25
Deferred - argument arity fixes for stricter Flow updates

pranaygp:
Closes #258

This is a duplicate of #258, but includes fixes for `Deferred#then` and `Deferred#catch`


zpao:
I lied and just folded your changes into a single commit in #258, which I needed to rebase to drop the lint change. So I'm going to close this here but I'll do a merge commit in other PR so you get the right blame :)

PR #262, opened 2017-07-26
Include built files for flow test

### LINKED ISSUE #261: Add test to ensure compiled Flow files are valid
We can avoid #260, so let's add a test of some sort.

pranaygp:
Closes #261

This should fail travis


zpao:
Looks like this works but I think we should do it in addition to the existing flow check, not replacing it.

pranaygp:
It should be testing both, the src and lib files

PR #266, opened 2017-08-29
Allow range of minor versions of `eslint-plugin-relay`

flarnie:
**what is the change?:**
Changed semver requirements of peerdep. on `eslint-plugin-relay` in
`eslint-config-fbjs`

**why make this change?:**
Trying to get CI passing on Draft.js, I had to update
`eslint-config-fbjs` due to other peerdep. requirements, and need the
latest version of `eslint-plugin-relay` to get a bug fix for node 4.x.
So I need to be able to use the latest `eslint-config-fbjs` and
`eslint-plugin-relay` together.

**test plan:**
I manually changed the `package.json` in my `node_modules` in the
Draft.js repo and ran `npm install` and it didn't throw.

**issue:**
Trying to get CI passing on Draft.js github - see
https://github.com/facebook/draft-js/pull/1349


packages/eslint-config-fbjs/package.json
```diff
@@ -26,6 +26,6 @@
     "eslint-plugin-flowtype": "^2.35.0",
     "eslint-plugin-jsx-a11y": "^6.0.2",
     "eslint-plugin-react": "^7.1.0",
-    "eslint-plugin-relay": "0.0.8"
+    "eslint-plugin-relay": ">=0.0.8 <=0.0.17"
```
zpao: Let's just make this `^0.0.8` and that way it'll pick up future `0.0.z` versions. The risk that that they make a breaking change in there but we can deal with that if it happens.

kassens: Sounds right.

SimenB:
Can this get a release? `eslint-plugin-relay`'s dependency on a dated graphql creates a ton of flow errors

packages/eslint-config-fbjs/package.json
SimenB: This is wrong, you have to do `~0.0.8`.

image
image

https://semver.npmjs.com/

SimenB: #271

PR #269, opened 2017-09-24
Change license BSD+Patents -> MIT

sophiebits:


abdulhannanali:
@zpao Wow! @zpao Does Facebook still use eslint-config-fbjs? I want to use it in my project without preparing for a bad time seems stable to me

PR #271, opened 2017-10-14
Set ~0.0.8 as minimum for peer dep of eslint-plugin-relay

SimenB:
See #266


zpao:
Looks like #266 got abandoned so let's ship this. Thanks!

SimenB:
Will it get a release as well? 🙂

zpao:
Published v2.0.1

PR #272, opened 2017-10-18
Added a more updated alternative to isomorphic-fetch.

lquixada:
isomorphic-fetch hasn't been updated for a while, so cross-fetch was created in order to provide a more updated, flexible and [fixed](https://github.com/matthew-andrews/isomorphic-fetch/issues/125) alternative to the community.

### DIFF
--- a/packages/fbjs/src/__forks__/fetch.js
+++ b/packages/fbjs/src/__forks__/fetch.js
@@ -9,11 +9,8 @@
 
 'use strict';
 
-// This hopefully supports the React Native case, which is already bringing along
-// its own fetch polyfill. That should exist on `global`. If that doesn't exist
-// then we'll try to polyfill, which might not work correctly in all environments.
-if (global.fetch) {
-  module.exports = global.fetch.bind(global);
-} else {
-  module.exports = require('isomorphic-fetch');
-}
+// Add a universal fetch polyfill (React Native compatible)
+require('cross-fetch/polyfill')
+
+// Make it importable
+module.exports = global.fetch.bind(global);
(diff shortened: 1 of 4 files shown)


packages/fbjs/src/__forks__/fetch.js
```diff
@@ -9,11 +9,8 @@
...
-// This hopefully supports the React Native case, which is already bringing along
-// its own fetch polyfill. That should exist on `global`. If that doesn't exist
-// then we'll try to polyfill, which might not work correctly in all environments.
-if (global.fetch) {
-  module.exports = global.fetch.bind(global);
-} else {
-  module.exports = require('isomorphic-fetch');
-}
+// Add a universal fetch polyfill (React Native compatible)
+require('cross-fetch/polyfill')
```
zpao: I never really wanted to use a polyfill here - requiring ideally shouldn't mutate global state.

Would it maybe make sense to to not require the polyfill here if global.fetch exists (basically what we had here before and only replace the `require('isomorphic-fetch')` call?

I actually forget which of our users are even using this. I think it's just relay-classic and I'm sure they aren't depending on it actually being a polyfill so it's probably ok. We were also already deferring to ReactNative's or the browser if in those environments without polyfilling. What do you think?

lquixada: Totally agree with not touching the global state. I only suggested the polyfill because requiring isomorphic-fetch do that automatically. So I wanted to do a 1:1 conversion.

However, for the sake of consistency, I think it's better to depend on one implementation, be it the native one or the third-party one. The code in question relies on both leading to the same problem polyfill faces as sindresorhus discusses [here](https://github.com/sindresorhus/ponyfill#how-are-ponyfills-better-than-polyfills).

For instance, if [fetch spec](https://fetch.spec.whatwg.org/) receives a new feature and it's deployed to the browser, the chosen fallback library might not follow that improvement or might take long to publish it.

So, my suggestion is to use only the native fetch (consistent but gets some environments uncovered), to use only a third-party library (consistent but adds more codebase) or to leave this decision to the users.

lquixada: hi @zpao! any news on this matter?

lquixada:
upgraded the isomorphic-fetch alternative to version 3.

PR #282, opened 2018-02-08
Add memoization to hyphenateStyleName

necolas:
I'm working on a patch for React that calls this function, and thought it may be worth adding memoization.

I didn't know about this function until today, and would also like to use it another project in place of [this dependency](https://github.com/rexxars/hyphenate-style-name/blob/master/index.js) which memoizes the result.

### DIFF
--- a/packages/fbjs/src/core/hyphenateStyleName.js
+++ b/packages/fbjs/src/core/hyphenateStyleName.js
@@ -11,6 +11,7 @@
 'use strict';
 
 const hyphenate = require('hyphenate');
+const memoizeStringOnly = require('memoizeStringOnly');
 
 const msPattern = /^ms-/;
 
@@ -34,4 +35,4 @@ function hyphenateStyleName(string) {
   return hyphenate(string).replace(msPattern, '-ms-');
 }
 
-module.exports = hyphenateStyleName;
+module.exports = memoizeStringOnly(hyphenateStyleName);


aweary:
>I didn't know about this function until today, and would also like to use it another project

@necolas keep in mind that `fbjs` is meant for Facebook projects, and depending on it may not be a good idea. From the README:

>If you are consuming the code here and you are not also a Facebook project, be prepared for a bad time. APIs may appear or disappear and we may not follow semver strictly, though we will do our best to. This library is being published with our use cases in mind and is not necessarily meant to be consumed by the broader public. In order for us to move fast and ship projects like React and Relay, we've made the decision to not support everybody. We probably won't take your feature requests unless they align with our needs. There will be overlap in functionality here and in other open source projects.

packages/fbjs/src/core/hyphenateStyleName.js
```diff
@@ -31,7 +33,10 @@ const msPattern = /^ms-/;
  * @return {string}
  */
 function hyphenateStyleName(string) {
-  return hyphenate(string).replace(msPattern, '-ms-');
+  if (memoized[string] === undefined) {
+    memoized[string] = hyphenate(string).replace(msPattern, '-ms-');
+  }
+  return memoized[string];
```
zpao: Practically this seems fine. But let's use [`memoizeStringOnly`](https://github.com/facebook/fbjs/blob/4369c7dfeb1ab0de27feb3e9245f599c8edd6a5c/packages/fbjs/src/core/memoizeStringOnly.js) (`module.exports = memoizeStringOnly(hyphenateStyleName)`). That'll also make sure we don't do the wrong thing for `'hasOwnProperty'` 😜

zpao:
I tried to be scary in the readme about using fbjs for other things, *[but it hasn't worked](https://www.npmjs.com/browse/depended/fbjs)*, so whatever. I've stuck with semver (well, <v1.0 rules) so it hasn't mattered but I'm likely to drop some files that none of our projects are using in 0.9. I'm not going to intentionally break anybody and it'll always happen in appropriate version bumps, but I also won't feel bad when it inevitably happens. That's my additional warning, but this method won't be going anywhere because I think React will be around for a while, so you should be fine 😀

necolas:
Yeah it's no problem in my case, `react-native-web` is the kind of project where I think it makes sense to use the same functions used in `react`/`react-dom`/`react-native` for similar operations. If only to align the results rather than share code (which isn't possible anymore with flat bundles).

packages/fbjs/src/core/hyphenateStyleName.js
necolas: ok!

PR #289, opened 2018-05-10
Update ua-parser-js

robertjd:
This dependency is being flagged by security scanners, see https://snyk.io/vuln/npm:ua-parser-js:20180227

P.S. It looks like my version of npm is changing some formatting in package.json


zpao:
Thanks for this but I went with #291 instead.

Issue #132, opened 2016-03-27
Does Promise.prototype.finally turn rejected promise to resolved in purpose?


adventure-yunfei:
I just using react-native which requires the project but got this issue.

Other promise implementations like well-known bluebird or the "promise" package itself have a ".finally" which won't change the promise state. I'm not sure why here it is different.

kassens:
Can you post some code that shows this behavior?

spicyj:
```
var result = thingThatFails()
  .finally((res) => { cleanUp(); return res; });
```

makes `result` a resolved promise, even if `thingThatFails()` returned a rejected one. It seems standard now for people to keep the resolution/rejection through a `finally` so we should probably switch; someone needs to look over our internal code to see if anything will break.

PR #293, opened 2018-06-07
Fix Promise.finally Non-standard behavior

liuqiang1357:
Standard behavior: [Promise.prototype.finally()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally)

Related issue: [#19490](https://github.com/facebook/react-native/issues/19490), [#132](https://github.com/facebook/fbjs/issues/132)

### DIFF
--- a/packages/fbjs/src/__forks__/Promise.native.js
+++ b/packages/fbjs/src/__forks__/Promise.native.js
@@ -18,8 +18,13 @@ require('promise/setimmediate/done');
 /**
  * Handle either fulfillment or rejection with the same callback.
  */
-Promise.prototype.finally = function(onSettled) {
-  return this.then(onSettled, onSettled);
-};
+Promise.prototype.finally = function(callback) {
+  return this.then(
+    value => Promise.resolve(callback()).then(() => value),
+    reason => Promise.resolve(callback()).then(() => {
+      throw reason;
+    })
+  );
+}
 
 module.exports = Promise;


danlugo92:
How come this isn't merged yet.

zpao:
@hramos - this was legit for RN. See https://github.com/facebook/fbjs/pull/95. If it's not longer needed, can you confirm? We can ship a new fbjs 0.8.x release so RN doesn't have to update to the newer releases.

hramos:
@yungsters ^

packages/fbjs/src/__forks__/Promise.native.js
```diff
@@ -18,8 +18,13 @@ require('promise/setimmediate/done');
...
-  return this.then(onSettled, onSettled);
-};
+Promise.prototype.finally = function(callback) {
+  return this.then(
+    value => Promise.resolve(callback()).then(() => value),
+    reason => Promise.resolve(callback()).then(() => {
+      throw reason;
+    })
+  );
+}
```
zpao: FWIWI: We can replace this whole definition with just a call to `require('promise/setimmediate/finally')`. https://github.com/then/promise/blob/7.3.1/src/finally.js

What you have is virtually identical, just reducing the amount of code we need to have written ourselves.

zpao:
@yungsters Does RN's runtime not have a native Promise impl yet? Feels like we should do that, then swap this module out for something that just returns the existing global.

yungsters:
React Native's version of JSC on Android is super old. On iOS, we delegate to iOS. 👎

zpao:
Ok, then this (or my recommendation) are effectively the same as the spec (minus some nuance). There's https://github.com/es-shims/Promise.prototype.finally/blob/master/implementation.js but that's looking pretty overkill, especially since we might not even have a native Promise to begin with. There are only a couple small differences between the old impl and this. Namely resolution/reject value.

Alternatively, you could ship the exact Promise polyfill you want in RN, and then the module here could just check for the global first. May lead to some duplicate packaging though since it'd be a runtime check.

As for rolling it out… obviously it could lead to behavior changes. I'd probably do an internal audit of `finally()` callsites and get an idea of impact.

Happy to run the release here when you're ready. RN is locked at a version so can happen on separate schedules.

indexzero:
@zpao is there an ETA for this to land? Agree with you that logistically it's safe since RN is locked at a specific version. Colleague of mine lost a day and a half to this in RN so the struggle in the community is still real. 

cc/ @hramos

grabbou:
Now that JSC on Android is updated, maybe we can continue with this PR? @yungsters

PR #295, opened 2018-06-21
Allow eslint-plugin-babel 5 in eslint-config-fbjs

thymikee:
There should not be any breaking changes, as ESLint 4 is already required 
https://github.com/babel/eslint-plugin-babel/releases/tag/v5.0.0


SimenB:
Fixed in #305

thymikee:
Thanks!

PR #298, opened 2018-07-24
Destructuring requires

mjesun:
Enables the inlining of expressions like `var x = require('foo').x`, which can come from a) an `import {x} from 'foo'`, b) `const {x} = require('foo')` or c) `const foo = require('foo'); const x = foo.x`. This will enable further optimizations when detecting dead `require`s.

I simplified the plugin by making use of other Babel utilities, so that we don't have to manually track variables and scopes, and we equally treat all inlinable expressions. Also fixed a bug where `var foo = require('foo'), bar = require('bar')` was not properly getting inlined.

Added tests to ensure that new cases are covered, made sure all existing ones still pass.

### DIFF
--- a/packages/babel-preset-fbjs/plugins/inline-requires.js
+++ b/packages/babel-preset-fbjs/plugins/inline-requires.js
@@ -9,144 +9,82 @@
 
 /**
  * This transform inlines top-level require(...) aliases with to enable lazy
- * loading of dependencies.
+ * loading of dependencies. It is able to inline both single references and
+ * child property references.
  *
- * Continuing with the example above, this replaces all references to `Foo` in
- * the module to `require('ModuleFoo')`.
+ * For instance:
+ *     var Foo = require('foo');
+ *     f(Foo);
+ *
+ * Will be transformed into:
+ *     f(require('foo'));
+ *
+ * When the assigment expression has a property access, it will be inlined too,
+ * keeping the property. For instance:
+ *     var Bar = require('foo').bar;
+ *     g(Bar);
+ *
+ * Will be transformed into:
+ *     g(require('foo').bar);
  */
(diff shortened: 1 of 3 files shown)


packages/babel-preset-fbjs/plugins/inline-requires.js
```diff
@@ -9,144 +9,82 @@
...
+        var declaratorPath = requireAlias(path) || requireMemberAlias(path);
+        var declarator = declaratorPath && declaratorPath.node;
 
-      /**
-       * Collect top-level require(...) aliases.
-       */
-      CallExpression: function(path, state) {
-        var node = path.node;
+        if (declarator) {
+          declaratorPath.scope.crawl();
```
davidaurelio: The only thing that I don’t understand is why you have to crawl here, and not …

packages/babel-preset-fbjs/plugins/inline-requires.js
```diff
@@ -9,144 +9,82 @@
...
+              'Cannot re-assign a require; name: ' + name + ', line: ' + line
+            );
           }
 
-          // Remove the declaration.
-          path.parentPath.parentPath.remove();
-          // And the associated binding in the scope.
-          path.scope.removeBinding(varName);
+          binding.referencePaths.forEach(ref => ref.replaceWith(init));
+          declaratorPath.remove();
```
davidaurelio: … and not here.

packages/babel-preset-fbjs/plugins/inline-requires.js
mjesun: We need to crawl because another plugin might have changed the scope already; for instance, the `import` / `export` Babel plugin. We need to have accurate variable binding, counting and referencing for the plugin to work.

PR #301, opened 2018-08-17
Remove location when cloning "require" calls

mjesun:
Due to the fact that the same `require` call is getting inlined multiple times, the Chrome Developer Tool is showing multiple breakpoints at the requires. While this is _technically_ correct, it creates confusion, so we'll remove all location data from the node that will be moved.

Tests were added to ensure the behavior is correct.


davidaurelio:
Fine with me, but we lose locations for all nodes, right?

PR #303, opened 2018-08-30
Upgrade babel-preset-fbjs to use Babel 7

jstejada:
babel-preset-fbjs should be using Babel 7. This only caused minimal changes in the output covered by the tests, which I updated.

This also includes an update to other packages that had transitive dependencies on Babel 6, and an update to flow-bin to 0.80.0 to fix an error that showed up when upgrading other packages.

If this looks good, I'll do a major version update on `babel-preset-fbjs` package itself

**Test Plan:**
- Ran `npm run fbjs:build`
- Ran `npm run fbjs:test`
- Ran `npm run fbjs:flow`
- Ran `npm run preset:test`
- Ran `npm i` inside `packages/babel-preset-fbjs`
- Ran `npm run test` inside `packages/babel-preset-fbjs`


mjesun:
LGTM; half of the builds in TravisCI failed though. I'd just re-schedule the build by pushing again.

PR #305, opened 2018-09-15
fix: allow babel-eslint 9 and eslint 5

SimenB:
No breaking changes from the perspective of this package

### DIFF
--- a/packages/eslint-config-fbjs/package.json
+++ b/packages/eslint-config-fbjs/package.json
@@ -19,8 +19,8 @@
     "test": "echo \"Error: no test specified\" && exit 1"
   },
   "peerDependencies": {
-    "babel-eslint": "^7.2.3 || ^8.0.0",
-    "eslint": "^4.2.0",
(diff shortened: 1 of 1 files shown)


mjesun:
@SimenB I guess it needs a release, right? I'll do that on Monday :)

SimenB:
Yes please!

Btw, any reason for https://github.com/facebook/jest/tree/master/packages/eslint-config-fb-strict to live in Jest? If we want those rules in Jest, we should just add it to the config in that repo. If we do want to keep the package, shouldn't it live in this repo?

SimenB:
@mjesun ping 😀

mjesun:
@SimenB I think @jstejada released it yesterday :)

SimenB:
Nope :( https://yarnpkg.com/en/package/eslint-config-fbjs

mjesun:
Released!

SimenB:
woo! I'll update in Jest, we should avoid the peer dep warnings

PR #308, opened 2018-09-19
Fix fbjs-scripts dependency on babel-preset-fbjs

jstejada:
#303 accidentally changed the dependency `fbjs-scripts` had on `babel-preset-fbjs` to be a relative file path. This breaks when installing `fbjs-scripts` from npm.

This commit fixes that


jstejada:
Thanks! @zpao going to merge this in and do a patch version release of `fbjs-scripts` if that sounds good to you

PR #309, opened 2018-09-26
In Babel 7, a preset should always export a function by default

jstejada:
When running babel-preset-env ^3.0.0, I'm running into the following error:

```
Plugin/Preset files are not allowed to export objects, only functions. In /Users/jstejada/relay/node_modules/babel-preset-fbjs/index.js
```

This PR makes it so `babel-preset-env` exports the configure function by default


jstejada:
Sounds good!

PR #310, opened 2018-09-27
Update changelog

jstejada:
It seems that the CHANGELOG on master is way behind the one on branch `fbjs-0.8.x`, so this commit updates it with the changes reflected in the `fbjs-0.8.x` branch.

Not sure what the setup between the 2 branches is. It looks like the `fbjs-0.8.x` branch only contains the `fbjs` package? Has the code from that branch been merged into master?

I was going to try to udpate the changelog, but I was confused as to what changes are in which branch, and if master was up to date with the changes from 0.8.x, or if they completely diverged.

### DIFF
--- a/packages/fbjs/CHANGELOG.md
+++ b/packages/fbjs/CHANGELOG.md
@@ -1,19 +1,79 @@
 ## [Unreleased]
 
-### Added
-- More modules, primarily focused on the requirements of Fixed Data Table.
+## [0.8.17] - 2018-06-11
 
-### Removed
-- Remove whitelisting of React PropType warning from `warning` module.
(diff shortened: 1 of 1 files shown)


packages/fbjs/CHANGELOG.md
```diff
@@ -1,20 +1,80 @@
 ## [Unreleased]
 
-### Added
-- More modules, primarily focused on the requirements of Fixed Data Table.
```
zpao: You shipped a 1.0 with all of this list of things you removed. Some of them may have shipped in the 0.8 releases (though definitely not this one). Please check though.

jstejada:
Will update the changelog with the changes for 1.0.0

PR #311, opened 2018-09-28
Add prettier config to package.json

kassens:


zpao:
Is this the same config we have internally? Don't want to have anything autoformatted that would make divergence hard to handle.

kassens:
The one small divergence is trailingCommas: es5 which is slightly less then `all`. This allows packages like SignedSource to not require a babel transform before publishing. I think it mostly affects trailing commas in function arguments.

package.json
```diff
@@ -5,18 +5,27 @@
...
+    "eslint-plugin-babel": "^4.1.1",
     "eslint-plugin-flowtype": "^2.35.0",
     "eslint-plugin-jsx-a11y": "^6.0.2",
     "eslint-plugin-react": "^7.1.0",
-    "eslint-plugin-relay": "0.0.8"
+    "eslint-plugin-relay": "0.0.8",
+    "prettier": "^1.14.3"
   },
   "scripts": {
+    "prettier": "find . -name node_modules -prune -or -name dist -prune -or -name '*.js' -print | xargs prettier --write",
```
zpao: I assume you ran this too and the only change was signed source?

kassens: Correct as we have `requirePragma: true`, otherwise there was a lot of reformatting that I at least wanted to do separately if desired.

PR #312, opened 2018-09-28
add .watchmanconfig

kassens:
This allows watchman (for jest) to run on this project on Facebook corporate managed machines.


kassens:
Assuming this isn't controversial as it only adds a tiny bit of dotfile clutter.

PR #317, opened 2018-10-11
Support `ignoredRequires` on `inline-requires` Plugin

yungsters:


packages/babel-preset-fbjs/plugins/__tests__/inline-requires-test.js
```diff
@@ -1,180 +1,79 @@
...
+      snapshot: true,
+    },
+
+    'inlines `Foo` before it is declared': {
+      code: "const doSomething = () => Foo; const Foo = require('Foo');",
+      snapshot: true,
+    },
+
+    'inlines `Foo` when required via `customRequire`': {
+      code: "const Baz = require('Bar').Baz; Baz;",
```
rafeca: To make this test match the description shouldn't the code be `const Baz = customRequire('Bar').Baz; Baz;`?

packages/babel-preset-fbjs/plugins/__tests__/__snapshots__/inline-requires-test.js.snap
```diff
@@ -0,0 +1,112 @@
...
+"
+`;
+
+exports[`inline-requires inlines \`Foo\` properties with destructuring: inlines \`Foo\` properties with destructuring 1`] = `
+"
+const {Baz} = require('Foo'); Baz;
+
+      ↓ ↓ ↓ ↓ ↓ ↓
+
+const { Baz } = require('Foo');Baz;
```
rafeca: This require should be inlined... 😕

mjesun: This is incorrect.

rafeca:
Based on the tests, it seems that the plugin does not inline destructured requires... that's weird...

Can you add a couple more tests?

- One that checks that removes a `require` if it's not used
- Another that checks that leaves a `require` if it's unassigned

packages/babel-preset-fbjs/plugins/__tests__/__snapshots__/inline-requires-test.js.snap
```diff
@@ -0,0 +1,112 @@
...
+"
+`;
+
+exports[`inline-requires inlines \`Foo\` when required via \`customRequire\`: inlines \`Foo\` when required via \`customRequire\` 1`] = `
+"
+const Baz = require('Bar').Baz; Baz;
+
+      ↓ ↓ ↓ ↓ ↓ ↓
+
+const Baz = require('Bar').Baz;Baz;
```
mjesun: This is incorrect too.

mjesun:
It looks like most of the cases the old tests covered were removed (e.g. integration with import commonjs interoperability, and property access). Could you add all of the existing back?

packages/babel-preset-fbjs/plugins/__tests__/__snapshots__/inline-requires-test.js.snap
mjesun: Note: that's per se _not_ incorrect, as we rely on the destructuring Babel transform to lower `const {a: c} = require('b')` to `const c = require('b').a`; but that's however not what the existing test asserted.

yungsters:
I hope the new version of this pull request is easier to follow. It only introduces two new tests, and all existing tests continue to pass.

PR #318, opened 2018-10-11
fix/polyfill performance.now for react-native

Esemesek:
Fixes https://github.com/facebook/react-native/issues/16151 in `react-native`


thymikee:
ping @cpojer you're now on RN team :D

cpojer:
Thank you for the fix. Would you mind adding a test for this as well?

cpojer:
I do not know about the release process and it will take a while to upgrade fbjs at FB (ugh) to sync out back to RN. Sorry about that.

cc @zpao: could you help release a new fbjs version? Thank you.

PR #330, opened 2019-01-13
fix: update dated peer deps in eslint config

SimenB:
I wanna bump these in Jest.

(note that a new major of `babel-eslint` is coming any day now as well)

### DIFF
--- a/packages/eslint-config-fbjs/package.json
+++ b/packages/eslint-config-fbjs/package.json
@@ -19,12 +19,12 @@
     "test": "echo \"Error: no test specified\" && exit 1"
   },
   "peerDependencies": {
-    "babel-eslint": "^7.2.3 || ^8.0.0 || ^9.0.0",
+    "babel-eslint": "^7.2.3 || ^8.0.0 || ^9.0.0 || ^10.0.0",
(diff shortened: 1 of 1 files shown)


packages/eslint-config-fbjs/package.json
```diff
@@ -19,12 +19,12 @@
     "test": "echo \"Error: no test specified\" && exit 1"
   },
   "peerDependencies": {
-    "babel-eslint": "^7.2.3 || ^8.0.0 || ^9.0.0",
+    "babel-eslint": "^7.2.3 || ^8.0.0 || ^9.0.0 || ^10.0.0",
```
SimenB: 11 is in beta, but it requires a babel config file. not sure if you wanna handle that somehow?

arcanis: Do you need the 11 for Jest, or would the 10 be enough for now?

SimenB: It's needed for TypeScript, but we'll probably go for https://github.com/typescript-eslint/typescript-eslint/ and not `babel-eslint`

SimenB:
Fixed by #343

PR #331, opened 2019-01-22
[eslint-config-fbjs] remove relay dependency

kassens:
The Relay eslint plugin is intended for product code, I don't think
it makes sense to pull in (likely outdated) GraphQL dependencies into
the framework projects using the fbjs eslint config.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,6 @@
     "eslint-plugin-flowtype": "^2.35.0",
     "eslint-plugin-jsx-a11y": "^6.0.2",
     "eslint-plugin-react": "^7.1.0",
-    "eslint-plugin-relay": "0.0.8",
     "prettier": "^1.14.3"
   },
   "scripts": {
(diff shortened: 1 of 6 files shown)


jstejada:
@kassens which framework projects are using this config? does external product code not use this config?

zpao:
I think the intent here was so anything we did opensource using Relay would just use this config and we wouldn't make a bunch of different configs. But I don't know that we ever open sourced anything using Relay (besides maybe Relay itself, eg tests?). Maybe Nuclide?

kassens:
Except for examples, it's hard for me to imagine publishing projects that use Relay since that relies on a server schema.

@jstejada for example Relay itself consumes this config, since eslint-plugin-relay is outdated, it pulls in `graphql@14.0.0` in addition to `graphql@14.1.0`.

I'll go ahead and merge this then as the reason seemed to have only been speculative. We can always add it back if we find ourself needing it.

PR #332, opened 2019-01-24
Update check-dependencies to read new format of `yarn outdated`

jrwats:
The field for `type` is 5 when workspaces are specified in the project. Let's just use the indices specified in `head`.

### DIFF
--- a/packages/fbjs-scripts/gulp/check-dependencies.js
+++ b/packages/fbjs-scripts/gulp/check-dependencies.js
@@ -17,6 +17,10 @@ var fancyLog = require('fancy-log');
 
 var PLUGIN_NAME = 'check-dependencies';
 
+const NAME = 0;
+const CURRENT = 1;
+const TYPE = 5;
+
 module.exports = function(opts) {
   function read(file, enc, cb) {
     var cwd = path.dirname(file.path);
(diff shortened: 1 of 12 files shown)


kassens:
Can you rebase on #333? That changed the CI config as well, I think with the same change.

jrwats:
Wow, maybe I'm dumb and it's actually node's version that matters?  I'm running v10.11.0 where my fix is "necessary".

packages/fbjs-scripts/gulp/check-dependencies.js
```diff
@@ -38,17 +38,27 @@ module.exports = function(opts) {
...
-          .map(d => JSON.parse(d))[1]['data']['body'];
+          .map(d => JSON.parse(d))
+          .filter(j => j.type === 'table')[0].data;
       } catch (e) {
         console.log('error', e)
         cb(new PluginError(PLUGIN_NAME, 'npm broke'));
       }
 
+      // Convert ["Package", "Current",...] to {"Package": 0, ...}
+      const name2idx = {};
```
zpao: > `name2idx`

ಠ_ಠ

jrwats: nameToIdx?

zpao:
Alright, rebased and squashed into a single commit: https://github.com/facebook/fbjs/commit/ccad8f1e81daa851d18cc84153b5e1d595f1edef

PR #333, opened 2019-01-24
Install latest yarn on TravisCI

zpao:
#332 is failing because Travis has an old version of Yarn (v1.3.2) installed. Output format has changed at some point (but semver?) so let's just make sure we're always running the latest.


kassens:
CI passes, so lgtm.

PR #334, opened 2019-01-25
Include "require.resolve" of a "require" as an inlineable call

mjesun:
In tests we sometimes use `require(require.resolve('module'))`, which is the same as `require('module')`, so we add an additional case to be sure both are inlined. Also added tests to verify the behavior.

### DIFF
--- a/packages/babel-preset-fbjs/plugins/inline-requires.js
+++ b/packages/babel-preset-fbjs/plugins/inline-requires.js
@@ -160,9 +160,10 @@ function isInlineableCall(node, state) {
     node['arguments'][0].type === 'CallExpression' &&
     node['arguments'][0].callee.type === 'MemberExpression' &&
     node['arguments'][0].callee.object.type === 'Identifier' &&
-    node['arguments'][0].callee.object.name === 'require' &&
+    state.inlineableCalls.hasOwnProperty(node['arguments'][0].callee.object.name) &&
     node['arguments'][0].callee.property.type === 'Identifier' &&
     node['arguments'][0].callee.property.name === 'resolve' &&
+    node['arguments'][0]['arguments'].length >= 1 &&
     node['arguments'][0]['arguments'][0].type === 'StringLiteral' &&
     !state.ignoredRequires.hasOwnProperty(node['arguments'][0]['arguments'][0].value);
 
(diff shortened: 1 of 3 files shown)


packages/babel-preset-fbjs/plugins/inline-requires.js
```diff
@@ -142,12 +142,29 @@ function inlineableMemberAlias(path, state) {
...
-  );
+    !state.ignoredRequires.hasOwnProperty(node['arguments'][0].value);
+
+  // require(require.resolve('foo'));
+  const isRequireResolveCall =
+    isInlineable &&
+    node['arguments'][0].type === 'CallExpression' &&
+    node['arguments'][0].callee.type === 'MemberExpression' &&
+    node['arguments'][0].callee.object.type === 'Identifier' &&
+    node['arguments'][0].callee.object.name === 'require' &&
```
rafeca: To keep things consistent we should also check if the callee name is in `inlineableCalls` (like we do in L148).

packages/babel-preset-fbjs/plugins/inline-requires.js
```diff
@@ -142,12 +142,29 @@ function inlineableMemberAlias(path, state) {
 }
 
 function isInlineableCall(node, state) {
-  return (
+  const isInlineable =
     node.type === 'CallExpression' &&
     node.callee.type === 'Identifier' &&
     state.inlineableCalls.hasOwnProperty(node.callee.name) &&
-    node['arguments'].length === 1 &&
+    node['arguments'].length >= 1;
```
rafeca: I guess it's fine if we also inline `require` calls with multiple arguments, can we add a simple test for this though?

packages/babel-preset-fbjs/plugins/inline-requires.js
```diff
@@ -142,12 +142,29 @@ function inlineableMemberAlias(path, state) {
...
+  // require(require.resolve('foo'));
+  const isRequireResolveCall =
+    isInlineable &&
+    node['arguments'][0].type === 'CallExpression' &&
+    node['arguments'][0].callee.type === 'MemberExpression' &&
+    node['arguments'][0].callee.object.type === 'Identifier' &&
+    node['arguments'][0].callee.object.name === 'require' &&
+    node['arguments'][0].callee.property.type === 'Identifier' &&
+    node['arguments'][0].callee.property.name === 'resolve' &&
+    node['arguments'][0]['arguments'][0].type === 'StringLiteral' &&
```
rafeca: Can you check that `require.resolve()` is called with >= 1 argument? even if it would be incorrect to call it without arguments we don't want to throw during the transpilation...

Issue #347, opened 2019-06-19
Invariant function throws error if __DEV__ is false instead of true


alihussain5:
In https://github.com/facebook/fbjs/pull/240/files, it looks like there was a change that turned

```javascript
var validateFormat = function(format) {};

if (__DEV__) {
  validateFormat = function(format) {
    if (format === undefined) {
      throw new Error('invariant requires an error message argument');	
    }	
  };	
}
```

into

```javascript
const validateFormat = __DEV__
  ? function(format: mixed): void {}
  : function(format: mixed): void {
      if (format === undefined) {
        throw new Error('invariant(...): Second argument must be a string.');
      }
    };
```

It looks like this change actually caused the logic to be flipped. It used to only throw an error `if __DEV__` but after the change it only throws an error `if !__DEV__`.

We use `relay-runtime` which uses `fbjs/invariant`. In our production environment, users kept getting an error of the form:

```
invariant(...): Second argument must be a string.
```

And I traced it down to this. They should've been getting an error of this form instead:

```
Minified exception occurred; use the non-minified dev environment for the full error message and additional helpful warnings.
```

Which had never had a chance to throw because the other error was being thrown first.

PR #348, opened 2019-06-19
Fix invariant function to throw error only if __DEV__ is true

alihussain5:
Related issue: https://github.com/facebook/fbjs/issues/347

A change was made in https://github.com/facebook/fbjs/pull/240/files that incorrectly flipped the logic to only throw an error if `__DEV__ === false`. This would change it back to only throw the error if `__DEV__ === true` as was the behaviour previously.


zpao:
Oops. Thanks!

PR #351, opened 2019-08-20
fix: allow eslint v6 as peer dependency

SimenB:
Avoids a peer dep warning for Jest.

@cpojer something you can get landed and released?


SimenB:
Effective! 😀

PR #353, opened 2019-09-22
#314 - Fix isSigned being inverted. Add test for re-signing an existing file.

inlineblock:
The isSigned is inverted as #314 noted. Fixing that and updating tests. This allow now makes the re-signing of files with signFile to not fail. Adding a test for that too.


kassens:
published as v2.0.0 due to the breaking change.

PR #356, opened 2019-10-11
inline-requires: Ignore modules with member assignmments

yungsters:
Currently, the `inline-requires` plugin breaks semantics when a property of a module is assigned a value. For example:

```
var X = require('X');
var origA = X.a;
X.a = function() {
  origA();
};
```

Currently, this will become:

```
require('X').a = function() {
  require('X').a();
};
```

Semantics are broken and this new code will loop infinitely. Instead, the intermediate `origA` alias must be preserved. This is the new output:

```
var origA = require('X').a;
require('X').a = function() {
  origA();
};
```

**Test Plan:**

Ran `yarn test` and verified Jest snapshots.


yungsters:
The Travis CI failures (running `yarn run test` from `packages/fbjs`) repros on master. I do not believe those failures are related to these changes.

Issue #252, opened 2017-07-13
setImmediate is broken in IE10 / IE11.


wmill:
Calls to setImmediate won't working in IE10 or IE11.

Known bug since 2014, https://github.com/YuzuJS/setImmediate/issues/35

Basically those versions of IE don't allow you to make a naked call to `setImmediate`, it must be called as `window.setImmediate`. However the setImmediate package doesn't install if there's a global setImmediate.

https://github.com/YuzuJS/setImmediate/blob/master/setImmediate.js#L4-L6

Workaround is found here, https://github.com/reflux/refluxjs/issues/374

Briefly just set `var setImmediate = window.setImmediate.bind(window);` if there's a global.setImmediate.

I know it's really a bug in a separate package, but it's been open since 2014. I'm hoping you have higher standards and will implement the workaround :)

Issue #290, opened 2018-05-19
"global is not defined" in setImmediate.js on v0.8.16


zhulin2609:

zhouhaoxiang:
I have the same problem

thomassuckow:
global got renamed to globalThis in the tc39 proposal

frenzzy:
probably to fix this issue need to change this line:
https://github.com/facebook/fbjs/blob/35d7693c0908820af5a0ed2635dbae45ab6feeb8/packages/fbjs/src/stubs/ErrorUtils.js#L12
to
```js
if (typeof global !== 'undefined' && global.ErrorUtils) {
```

Stupidism:
Any news?

shiyun:
> I have the same problem

Me too!

LukeXF:
Still an issue, any update?

joriewong:
Same issue

k-kyler:
This worked for me.

<script>
      const global = globalThis;
</script>

lamasagar:
> This worked for me.
> 
> <script> const global = globalThis; </script>

This worked but can any one explain what this is doing? 
is there a way to place this inside the react component?

thamizh-qwerty:
> This worked for me.
> 
> <script> const global = globalThis; </script>

It does not works in angular + electronJS + contextIsolation = true

PR #387, opened 2020-05-20
#252 Make "setImmediate" browser compatible

ptolomeus:
Fix exported `setImmediate` polyfill, which was broken especially in IE11 (and maybe also earlier), which has actually supports it natively.


ptolomeus:
Fixes also #290

jbis9051:
Any update?

PR #403, opened 2020-10-07
Remove `core-js` from `fbjs` and `fbjs-scripts`

cpojer:
This drops `core-js` as a dependency from fbjs and pushes the responsibility to add polyfills for Map and Set to the consumer of fbjs. Since fbjs is not actively used outside of Facebook, and ES2015+ has been implemented in all browsers, I feel confident this change will not lead to major headaches for most people. This also aligns fbjs with React which does not ship with polyfills and expects users to provide them.

## Changes:
* `fbjs`: `core-js` is used for Map and Set. I removed those modules and am relying on the native Set and Map implementations instead.
* `fbjs-scripts` adds `core-js` for Jest but all Node.js versions that Jest runs on support the necessary language features.

Note, `core-js` is almost 8 MiB but we only pull in very little code.

## Test Plan
* Ran the JavaScript tests, they pass.


yungsters (approved):

zpao:
For fbjs-scripts 3, we probably want to remove these since they are not in fbjs anymore: https://github.com/facebook/fbjs/blob/0de589c70c0248ec093ca8e35b6aa6bd16fcac5c/packages/fbjs-scripts/third-party-module-map.json#L2-L5. Can you ship a patch version with that? Shouldn't be overly detrimental as is since consumers aren't typically using those requires directly but best to be accurate. If they are, moving forward they would need to do that mapping themselves.

cpojer:
Yap, noticed it yesterday: https://github.com/facebook/fbjs/pull/404

PR #405, opened 2020-10-26
chore: move eslint-config-fb-strict into this repo

SimenB:
This currently lives in the Jest repo. I don't think there's a good reason for that, especially as that repo has moved away from Flow and `babel-eslint` towards a TS toolchain.

I plan to delete the source code from Jest's repo

https://github.com/facebook/jest/tree/92f74cd1fa057bc71890cd01018e6287fd24612b/packages/eslint-config-fb-strict


cpojer (approved):
👍

PR #406, opened 2020-11-05
Fix require shadowing in `inline-requires` plugin

namaggarwal:
1. It's possible that the require is overridden in a module (or scope) and no longer points to the CommonJS require function. The plugin doesn't respect this. For this, we will now honor the require redeclaration if in global scope.
2. If the require function is re-declared in a sub-scope, e.g. a function. The plugin changes the semantics of the program by inlining the require call without renaming the scoped require declaration (it then accidentally calls the custom require implementation). This is fixed by renaming local scope "require" functions and calls to "_require"

### DIFF
--- a/packages/babel-preset-fbjs/plugins/__tests__/__snapshots__/inline-requires-test.js.snap
+++ b/packages/babel-preset-fbjs/plugins/__tests__/__snapshots__/inline-requires-test.js.snap
@@ -10,6 +10,32 @@ var foo = require(\\"foo\\");
 "
 `;
 
+exports[`inline-requires does not transform require calls if its not needed: does not transform require calls if its not needed 1`] = `
+"
+function test () {
+  function require(condition) {
+    if (!condition) {
+      throw new Error('The condition is false');
+    }
+  }
+
+  require('test');
+}
+
+      ↓ ↓ ↓ ↓ ↓ ↓
+
+function test() {
+  function require(condition) {
+    if (!condition) {
+      throw new Error('The condition is false');
+    }
+  }
+
+  require('test');
+}
+"
+`;
+
 exports[`inline-requires ensures that the inlined require still points to the global require function even if local require is not called: ensures that the inlined require still points to the global require function even if local require is not called 1`] = `
 "
 const foo = require('foo');
@@ -34,6 +60,34 @@ function test() {
   }
 
   require('foo').isOnline();
+}
+"
+`;
+
+exports[`inline-requires ensures that the inlined require still points to the global require function with inlineableCalls options: ensures that the inlined require still points to the global require function with inlineableCalls options 1`] = `
+"
+const foo = customStuff('foo');
+
+function test() {
+  function customStuff(condition) {
+    if (!condition) {
+      throw new Error('Condition is falsy');
+    }
+  }
+
+  customStuff(foo.isOnline());
+}
+
+      ↓ ↓ ↓ ↓ ↓ ↓
+
+function test() {
+  function _customStuff(condition) {
+    if (!condition) {
+      throw new Error('Condition is falsy');
+    }
+  }
+
+  _customStuff(customStuff('foo').isOnline());
 }
 "
 `;
--- a/packages/babel-preset-fbjs/plugins/__tests__/__snapshots__/inline-requires-test.js.snap
+++ b/packages/babel-preset-fbjs/plugins/__tests__/__snapshots__/inline-requires-test.js.snap
@@ -10,7 +10,7 @@ var foo = require(\\"foo\\");
 "
 `;
 
-exports[`inline-requires does not transform require calls if its not needed: does not transform require calls if its not needed 1`] = `
+exports[`inline-requires does not transform require calls if it is not needed: does not transform require calls if it is not needed 1`] = `
 "
 function test () {
   function require(condition) {
--- a/packages/babel-preset-fbjs/plugins/__tests__/__snapshots__/inline-requires-test.js.snap
+++ b/packages/babel-preset-fbjs/plugins/__tests__/__snapshots__/inline-requires-test.js.snap
@@ -10,6 +10,62 @@ var foo = require(\\"foo\\");
 "
 `;
 
+exports[`inline-requires ensures that the inlined require still points to the global require function even if local require is not called: ensures that the inlined require still points to the global require function even if local require is not called 1`] = `
+"
+const foo = require('foo');
+
+function test() {
+  function require(condition) {
+    if (!condition) {
+      throw new Error('Condition is falsy');
+    }
+  }
+
+  foo.isOnline();
+}
+
+      ↓ ↓ ↓ ↓ ↓ ↓
+
+function test() {
+  function _require(condition) {
+    if (!condition) {
+      throw new Error('Condition is falsy');
+    }
+  }
+
+  require('foo').isOnline();
+}
+"
+`;
+
+exports[`inline-requires ensures that the inlined require still points to the global require function: ensures that the inlined require still points to the global require function 1`] = `
+"
+const foo = require('foo');
+
+function test() {
+  function require(condition) {
+    if (!condition) {
+      throw new Error('Condition is falsy');
+    }
+  }
+
+  require(foo.isOnline());
+}
+
+      ↓ ↓ ↓ ↓ ↓ ↓
+
+function test() {
+  function _require(condition) {
+    if (!condition) {
+      throw new Error('Condition is falsy');
+    }
+  }
+
+  _require(require('foo').isOnline());
+}
+"
+`;
+
 exports[`inline-requires ignores require properties (as identifiers) that are re-assigned: ignores require properties (as identifiers) that are re-assigned 1`] = `
 "
 var X = require(\\"X\\");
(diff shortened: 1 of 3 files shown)


packages/babel-preset-fbjs/plugins/__tests__/inline-requires-test.js
```diff
@@ -53,6 +53,56 @@ pluginTester({
       snapshot: false,
     },
 
+    'ensures that the inlined require still points to the global require function': {
```
MichaReiser, line 56: I probably haven't formatted the files correctly in my patch. Could you make sure that the file conforms to the formatting guidelines (or run prettier?).

packages/babel-preset-fbjs/plugins/inline-requires.js
```diff
@@ -68,8 +77,14 @@ module.exports = babel => ({
...
               }
-              const {declarationPath, moduleName} = parseResult;
 
+              // Do not transform require calls if require is redeclared in global scope
+              const requireBinding = path.scope.getBinding('require');
+              if (requireBinding != null && requireBinding.scope.parent == null) {
+                return;
+              }
+
+              const {declarationPath, moduleName} = parseResult;
```
MichaReiser, line 87: I don't think we should limit this check to the `global` scope only. E.g. we shouldn't perform any modification whenever a user calls a custom `require` function. E.g. doing any changes wouldn't be desired in the following example (worth adding a test case if it's missing). 

```
function test () {
  function require(condition) {
    if (!condition) {
      throw new Error(`The condition is false`);
    }
  }

  require('test');
}
```

There's another complexity that on top of this. The plugin allows consumers to pass custom `require` function names in `state.inlineableCalls`. Metro depends on this logic because it first lowers `import * from 'x'` and `import X from 'x'x` to `const x = importAll('x')` and `const X = importDefault('x')` calls that should then be inlined by this inline plugin (yes, this is not 100% spec conform). Metro will use unique names for the `importAll` and `importDefault` functions so it's technically not a problem but we shouldn't make assumptions in this plugin on how it's used.

That means that you can't test for a `require` bindings only but instead must test for the function name used by this specific `CallExpression`. That probably means that you'll have to extend `getInlineableModule` to return the `requireFunctionName` to have it available here.

packages/babel-preset-fbjs/plugins/inline-requires.js
```diff
@@ -39,6 +39,15 @@
...
   visitor: {
+    CallExpression(path, state) {
+      const node = path.node;
+      // Rename an existing require function to make sure any inserted `require` calls point to the
+      // global CommonJS require function.
+      const requireBinding = path.scope.getBinding('require');
+      if (requireBinding != null && requireBinding.scope.parent != null) {
+        path.scope.rename('require');
+      }
+    },
```
MichaReiser, line 50: This has the undesired side effect that you might be renaming `require` functions even if the file isn't using any `require` call. Plugins should aim to minimize changes to the input source. I would further argue that this change would already guarantee that you never have a `global` `require` function declared since all are renamed. 

I suggest that we move the renaming to the place where we perform the `inlining`. Before you inline a call, rename `require` in the current scope to `_require` (or `importDefault` to `_importDefault`, see my comment below`. and then inline the call.

MichaReiser (changes requested):
Thank you @namaggarwal  for working on this fix. We're heading in the right direction but need to take into account that the plugin allows custom require function names and we should try to avoid making any changes to the source file if it isn't strictly necessary.

packages/babel-preset-fbjs/plugins/__tests__/inline-requires-test.js
namaggarwal: I disabled max-len for this entire file.

packages/babel-preset-fbjs/plugins/__tests__/__snapshots__/inline-requires-test.js.snap
```diff
@@ -10,6 +10,116 @@ var foo = require(\\"foo\\");
...
+const foo = require('foo');
+
+function test() {
+  function require(condition) {
+    if (!condition) {
+      throw new Error('Condition is falsy');
+    }
+  }
+
+  foo.isOnline();
```
MichaReiser, line 50: Should this be 

```suggestion
  require(foo.isOnline());
```

To match the same test with `customStuff` and cover the case that the call-site of the "custom" `require` function is renamed?

I can see that the last test in this snapshot does exactly that. Do we need both tests?

packages/babel-preset-fbjs/plugins/__tests__/inline-requires-test.js
```diff
@@ -53,6 +54,103 @@ pluginTester({
...
+
+        function test() {
+          const foo = require('foo');
+          console.log(foo.test);
+        }
+      `,
+      snapshot: false,
+    },
+
+    'does not transform require calls if its not needed': {
```
MichaReiser, line 139: Nit
```suggestion
    'does not transform require calls if it is not needed': {
```

packages/babel-preset-fbjs/plugins/inline-requires.js
```diff
@@ -90,6 +97,12 @@ module.exports = babel => ({
               for (const referencePath of binding.referencePaths) {
                 excludeMemberAssignment(moduleName, referencePath, state);
                 try {
+                  const requireBinding = referencePath.scope.getBinding(
+                    requireName
+                  );
+                  if (requireBinding != null) {
+                    requireBinding.scope.rename(requireName);
+                  }
```
MichaReiser, line 105: The check for an existing `requireBinding` shouldn't be necessary. `referencePath.scope.rename` is a no-op if the existing name has no references.

```suggestion
                  referencePath.scope.rename(requireName);
```

packages/babel-preset-fbjs/plugins/inline-requires.js
```diff
@@ -107,7 +120,7 @@ module.exports = babel => ({
             ignoredRequires,
             inlineableCalls,
             membersAssigned: new Map(),
-          },
+          }
```
MichaReiser, line 123: Nit: Keep the comma

packages/babel-preset-fbjs/plugins/inline-requires.js
```diff
@@ -69,6 +69,13 @@ module.exports = babel => ({
                 return;
               }
               const {declarationPath, moduleName} = parseResult;
+              const requireName = path.node.callee.name || path.node.callee.property.name;
+
+              // If declaration path has require binding in the scope then
+              // don't inline calls
+              if(declarationPath.scope.getBinding(requireName) != null) {
+                return;
+              }
```
MichaReiser, line 78: I suggest to move the logic for getting the `requireName` and testing if there's no override in the current or any parent scope into the `getInlineableModule` function

```
function getInlineableModule(path, state) {
  const node = path.node;
  const fnName = node.callee.name;

  const isInlineable =
    node.type === 'CallExpression' &&
    node.callee.type === 'Identifier' &&
    state.inlineableCalls.has(fnName) &&
    node['arguments'].length >= 1 && 
    path.scope.getBinding(fnName) == null;

  if (!isInlineable) {
    return null;
  }
  
  // ....

  return moduleName == null || state.ignoredRequires.has(moduleName)
    ? null
    : { moduleName, requireFnName: fnName };
}
```

You'll have to rewrite the two call-sites to pass a `path` instead of a `node` and to get the `moduleName` and `requireFnName` from the returned object.

MichaReiser (changes requested):
Thank you for addressing comments. Looks pretty good now. I've to ask you for a few small changes but we should then be good to go 🥳

packages/babel-preset-fbjs/plugins/__tests__/__snapshots__/inline-requires-test.js.snap
namaggarwal: Yes, this is intentional and a different test as I wanted to test that it shouldn't rename the "require" function unnecessary

packages/babel-preset-fbjs/plugins/inline-requires.js
namaggarwal: Done. Can you please check the updated code

packages/babel-preset-fbjs/plugins/inline-requires.js
```diff
@@ -235,7 +249,21 @@ function getInlineableModule(node, state) {
...
+  let parentPath = path.parentPath;
+  let isRequireInScope = false;
+  const fnName = node.callee.name;
+  while (parentPath) {
+    if (parentPath.scope.getBinding(fnName) != null) {
+      isRequireInScope = true;
+      break;
+    }
+    parentPath = parentPath.parentPath;
+  }
```
MichaReiser, line 262: You don't have to traverse up in the tree. `scope.getBinding()` checks the current scope and any parent scope

https://github.com/babel/babel/blob/master/packages/babel-traverse/src/scope/index.js#L993-L1017

MichaReiser (commented):
I've a final small change to request before merging this.

packages/babel-preset-fbjs/plugins/inline-requires.js
namaggarwal: Thank you. Updated

MichaReiser (approved):

PR #407, opened 2020-11-12
Eliminate usage of @flow weak in keyMirrorRecursive

gkz:
`@flow weak` is deprecated. Replace single instance with `@flow` instead and add suppression on error which was previously hidden.


bigfootjon:
This has been superseded by https://github.com/facebook/fbjs/pull/469

PR #413, opened 2021-01-05
Bump "promise" to ^8.0.3 to match React Native

bengourley:
**tl;dr** out of sync dependencies on "promise" in React Native vs. fbjs mean that two different versions get installed on a project, and unhandled promise rejections do not get reported in Bugsnag on RN 0.63+ and Expo SDK39+

---

React Native 0.63 (and by extension Expo SDK 39-40) bumped its dependency on "promise" to be ^8.0.3 ¹

This repo depends on promise@^7.1.1 which has no valid overlapping versions. This module "fbjs" is depended on by "metro", which in turn is depended on by RN/Expo projects.

This means that on a standard RN/Expo project, two versions of "promise" get installed, making it impossible to obtain a reference to the same version that React Native is using², and thus impossible to detect unhandled promise rejections – and for example report them to Bugsnag.

¹ https://github.com/facebook/react-native/commit/b23efc526446cad53adc3de7465fd8be7309d84e
² https://github.com/bugsnag/bugsnag-js/blob/eb6e05f350f5acf496c9e3dfd341c9061c400c1f/packages/plugin-react-native-unhandled-rejection/rejection-handler.js#L5

### DIFF
--- a/packages/fbjs/package.json
+++ b/packages/fbjs/package.json
@@ -67,7 +67,7 @@
     "fbjs-css-vars": "^1.0.0",
     "loose-envify": "^1.0.0",
     "object-assign": "^4.1.0",
-    "promise": "^7.1.1",
+    "promise": "^8.0.3",
     "setimmediate": "^1.0.5",
     "ua-parser-js": "^0.7.18"
   },
--- a/packages/fbjs/yarn.lock
+++ b/packages/fbjs/yarn.lock
@@ -696,7 +696,7 @@ arrify@^1.0.0, arrify@^1.0.1:
   resolved "https://registry.yarnpkg.com/arrify/-/arrify-1.0.1.tgz#898508da2226f380df904728456849c1501a4b0d"
   integrity sha1-iYUI2iIm84DfkEcoRWhJwVAaSw0=
 
-asap@~2.0.3:
+asap@~2.0.6:
   version "2.0.6"
   resolved "https://registry.yarnpkg.com/asap/-/asap-2.0.6.tgz#e50347611d7e690943208bbdafebcbc2fb866d46"
   integrity sha1-5QNHYR1+aQlDIIu9r+vLwvuGbUY=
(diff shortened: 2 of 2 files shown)


jer-sen:
could this PR be merged ?

tkieft:
Any hope of getting this merged? We just ran into this in our project...

zpao:
React Native stopped using fbjs a year ago (https://github.com/facebook/react-native/commit/54e19a6b7f217ffc0611e660f2a6b1a8ad14775b), as did metro (https://github.com/facebook/metro/commit/203fb31ec028f9f147bd0a1b689b587a396d9b37). I know RN has done a release or 2 since then so I'd recommend upgrading RN as a better approach. Merging and shipping a new current fbjs release doesn't help, so this needs to be applied to a 3 year old release of fbjs, which may have other consequences.

tkieft:
Great, so it looks like the fix is to update to a version of RN >= 0.64.0 (which transitively includes metro >= 0.64.0). Thanks!

jer-sen:
@zpao `react-native-web` still relies on `fbjs` which cause me trouble. Why can't you merge this PR?

jer-sen:
@zpao I use `sentry-expo` for bug tracking (as a LOT of people using Expo since its the recommanded and only supported bug tracker for managed workflow: https://docs.expo.dev/guides/using-sentry/). It needs to hook on `Promise` global object to report failed or unhandled promised. To do so it relies on `promise` module but it needs to use the same version than the one that hooked on `Promise` global object, otherwise Sentry do not received events. React Native depends on `promise: "^8.0.3"` but 2 other modules depend inderectly on an other version of `promise` through `fbjs`: `react-native-web#fbjs#promise` & `expo#fbemitter#fbjs#promise`. If `fbjs` dependency on `promise` were bumped to `^8.0.3` there would be only one version in my project and everything would work fine (as it does with this addition to my `package.json`: `"resolutions": { "react-native-web/**/promise": "^8.0.3", "expo/**/promise": "^8.0.3" },`).

So, could you merge this PR ?

SimenB:
Thoughts on landing this? Just spent a couple of hours debugging a maximum call stack error that in the end was resolved by removing our `core-js` polyfill of `Promise.allSettled` and adding a `resolutions` for `promise` to force v8 (and disabling Sentry's polyfilling). As with above, v7 from `fbjs` was the one hoisted and the one `@sentry/react-native` ended up pulling in.

https://github.com/getsentry/sentry-react-native/pull/1984

PR #466, opened 2021-10-22
Update CI Action

zpao:
- Update list of current & lts releases to test with
- Update to setup-node@v2 (with built in caching between versions in the
  matrix)


bigfootjon (approved):
LGTM

PR #469, opened 2021-12-11
Fix @flow weak usage

bigfootjon:
Flow v0.167 bans "@flow weak". Let's upgrade this module

### DIFF
--- a/packages/fbjs-scripts/package.json
+++ b/packages/fbjs-scripts/package.json
@@ -5,7 +5,7 @@
   "repository": "facebook/fbjs",
   "license": "MIT",
   "dependencies": {
-    "@babel/core": "^7.0.0",
+    "@babel/core": "^7.1.5",
     "ansi-colors": "^1.0.1",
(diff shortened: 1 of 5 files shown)


mroch:
Maybe we can get back in sync with the internal version: html/shared/fbjs/key-mirror/keyMirrorRecursive.js

I'm on my phone so didn't do a detailed comparison though

zpao:
Yea, things are dramatically out of date here. I've made a couple feeble attempts to get the actual module needs of our projects still using fbjs to trim & update but nothing successful. And we'd need to redo the whole build process too since things have changed so much.

bigfootjon:
I just want Flow to keep passing for fbjs (I'm aware I'm adding to the problem).

Realistically, we should FBShipIt this project from WWW to solve this, but that takes more effort than I have bandwidth for

bigfootjon:
@zpao @mroch looks like we can't just use the latest version from WWW. Any thoughts on using the reverting to the original commit I opened this PR with?

PR #487, opened 2022-07-08
feat: update eslint deps

bradzacher:
- update to ESLint v8
- switch babel-eslint for hermes-eslint
- update eslint plugin versions

### DIFF
--- a/packages/eslint-config-fbjs-opensource/package.json
+++ b/packages/eslint-config-fbjs-opensource/package.json
@@ -1,6 +1,6 @@
 {
   "name": "eslint-config-fbjs-opensource",
-  "version": "2.0.1",
+  "version": "3.0.0",
   "description": "",
   "author": "",
   "repository": "facebook/fbjs",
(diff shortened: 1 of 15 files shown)


packages/eslint-config-fbjs-opensource/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "eslint-config-fbjs-opensource",
-  "version": "2.0.1",
+  "version": "3.0.0",
```
zpao, line 3: Personally I prefer specific commits that bump versions, not in PRs (so we can land safely then release)

packages/eslint-config-fbjs/README.md
```diff
@@ -16,23 +16,23 @@ There are several exceptions:
 npm install --save-dev \
   eslint-config-fbjs \
   eslint-plugin-babel \
-  eslint-plugin-flowtype \
+  eslint-plugin-ft-flow \
   eslint-plugin-jsx-a11y \
   eslint-plugin-react \
```
zpao, line 21: theoretically, we probably want to add https://www.npmjs.com/package/eslint-plugin-react-hooks to this config (and make sure the right rules are enabled) but that's distinct from the effort you have now. Fits with my version bump outside PRs though :)

zpao (changes requested):

PR #521, opened 2023-10-27
chore(deps): remove `cross-fetch` dependency

### LINKED ISSUE #520: Consider removing `cross-fetch` dependency
Since the Fetch API has been marked [as stable in Node.js 21](https://nodejs.org/dist/latest-v21.x/docs/api/globals.html#fetch), and is available without a flag in all actively maintained LTS versions (v18, v20), it might be time to remove the `cross-fetch` dependency.

jonkoops:
Removes the `cross-fetch` dependency from the package, as the Fetch API is now supported in Node.js.

Closes #520


packages/fbjs/package.json
```diff
@@ -63,7 +63,6 @@
     ]
   },
   "dependencies": {
-    "cross-fetch": "^3.1.5",
```
jonkoops, line 66: Side-note, I am not sure why this was included as a dependency, it seems like this was only ever used during testing.

zpao: fbjs was created to support FB projects, goal being to provide the same core/shared libraries we had internally. `fetch` was one of those things but had too much internal-specific code so it was a "forked" implementation here. It relied on `cross-fetch` and re-exported it (as you saw) but was not otherwise used directly.

